Issue #202, opened 2022-01-19
[FEATURE] Implement Circuit Breaker in Opensearch-js client


yilintao-amzn:
**Is your feature request related to a problem?**
Yes. When customer's query response is quite large, the opensearch-dashboard could have JS run out of heap issue.

**What solution would you like?**
Implement a circuit breaker in the opensearch-js client to detect the size of the return response data. If the size exceeds the threshold, it will trigger the circuit breaker.

**What alternatives have you considered?**
Increase the memory assigned to the opensearch-dashboards, but the issue of JS run out of heap still exists.

**Do you have any additional context?**
No.

PR #207, opened 2022-03-11
[Circuit-Breaker] Add memory threshold in opensearch-js client

zuochengding:
Signed-off-by: Zuocheng Ding <zding817@gmail.com>

### Description
This is to add memory checker to avoid dashboard out of memory crash.
There will be one field introduced in `ClientOptions`:  `memoryCircuitBreaker`. 
The default value of `memoryCircuitBreaker.enabled` is false until customer add manual config to modify it.
The default value of `memoryCircuitBreaker.maxPercentage` is `1.0` until customer modify it. The proper values should be within [0, 1] inclusively. 

 
### Issues Resolved
https://github.com/opensearch-project/opensearch-js/issues/202
 
### Check List
  - [X] All tests pass

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -126,6 +126,10 @@ interface ClientOptions {
     password?: string;
   };
   disablePrototypePoisoningProtection?: boolean | 'proto' | 'constructor';
+  memoryCircuitBreaker?: {
+    enabled: boolean;
+    maxPercentage: number;
+  }
 }
 
 declare class Client {
--- a/index.js
+++ b/index.js
@@ -178,7 +178,8 @@ class Client extends OpenSearchAPI {
       generateRequestId: options.generateRequestId,
       name: options.name,
       opaqueIdPrefix: options.opaqueIdPrefix,
-      context: options.context
+      context: options.context,
+      memoryCircuitBreaker: options.memoryCircuitBreaker
     })
 
     this.helpers = new Helpers({
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -51,6 +51,7 @@ const clientVersion = require('../package.json').version
 const userAgent = `opensearch-js/${clientVersion} (${os.platform()} ${os.release()}-${os.arch()}; Node.js ${process.version})`
 const MAX_BUFFER_LENGTH = buffer.constants.MAX_LENGTH
 const MAX_STRING_LENGTH = buffer.constants.MAX_STRING_LENGTH
+const HEAP_SIZE_LIMIT = require('v8').getHeapStatistics().heap_size_limit
 const kCompatibleCheck = Symbol('compatible check')
 const kApiVersioning = Symbol('api versioning')
 
@@ -81,6 +82,7 @@ class Transport {
     this.opaqueIdPrefix = opts.opaqueIdPrefix
     this[kCompatibleCheck] = 0 // 0 = to be checked, 1 = checking, 2 = checked-ok, 3 checked-notok
     this[kApiVersioning] = process.env.OPENSEARCH_CLIENT_APIVERSIONING === 'true'
+    this.memoryCircuitBreaker = opts.memoryCircuitBreaker
 
     this.nodeFilter = opts.nodeFilter || defaultNodeFilter
     if (typeof opts.nodeSelector === 'function') {
@@ -247,10 +249,10 @@ class Transport {
 
       const contentEncoding = (result.headers['content-encoding'] || '').toLowerCase()
       const isCompressed = contentEncoding.indexOf('gzip') > -1 || contentEncoding.indexOf('deflate') > -1
-
       /* istanbul ignore else */
       if (result.headers['content-length'] !== undefined) {
         const contentLength = Number(result.headers['content-length'])
+        // nodeJS data type limit check
         if (isCompressed && contentLength > MAX_BUFFER_LENGTH) {
           response.destroy()
           return onConnectionError(
@@ -260,6 +262,12 @@ class Transport {
           response.destroy()
           return onConnectionError(
             new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed string (${MAX_STRING_LENGTH})`, result)
+          )
+        } else if (shouldApplyCircuitBreaker(contentLength)) {
+          // Abort this response to avoid OOM crash of dashboards.
+          response.destroy()
+          return onConnectionError(
+            new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed heap memory limit.`, result)
           )
         }
       }
@@ -300,6 +308,13 @@ class Transport {
       response.on('error', onEnd)
       response.on('end', onEnd)
       response.on('aborted', onAbort)
+    }
+    // Helper function to check if memory circuit breaker enabled and the response payload is too large to fit into available heap memory.
+    const shouldApplyCircuitBreaker = (contentLength) => {
+      if (!this.memoryCircuitBreaker || !this.memoryCircuitBreaker.enabled) return false
+      const maxPercentage = validateMemoryPercentage(this.memoryCircuitBreaker.maxPercentage)
+      const heapUsed = process.memoryUsage().heapUsed
+      return contentLength + heapUsed > HEAP_SIZE_LIMIT * maxPercentage
     }
 
     const onBody = (err, payload) => {
@@ -634,6 +649,11 @@ function lowerCaseHeaders (oldHeaders) {
     newHeaders[header.toLowerCase()] = oldHeaders[header]
   }
   return newHeaders
+}
+
+function validateMemoryPercentage (percentage) {
+  if (percentage < 0 || percentage > 1) return 1.0
+  return percentage
 }
 
 module.exports = Transport
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "./": "./"
   },
   "homepage": "https://www.opensearch.org/",
-  "version": "1.0.2",
+  "version": "1.1.0",
   "versionCanary": "7.10.0-canary.6",
   "keywords": [
     "opensearch",
--- a/test/unit/client.test.js
+++ b/test/unit/client.test.js
@@ -951,6 +951,52 @@ test('Content length too big (string)', t => {
   })
 })
 
+test('Content length exceeds max heap limit', t => {
+  t.plan(4)
+  const percentage = 0.8
+  const HEAP_SIZE_LIMIT_WITH_BUFFER = Number(require('v8').getHeapStatistics().heap_size_limit * percentage)
+  const contentLength = buffer.constants.MAX_STRING_LENGTH - 1
+  const memoryCircuitBreaker = {
+    enabled: true,
+    maxPercentage: percentage
+  }
+  // Simulate allocation of bytes
+  const memoryAllocations = []
+  while (process.memoryUsage().heapUsed + contentLength <= HEAP_SIZE_LIMIT_WITH_BUFFER) {
+    const allocation = 50 * 1024 * 1024 // 50MB
+    const numbers = allocation / 8
+    const arr = []
+    arr.length = numbers
+    for (let i = 0; i < numbers; i++) {
+      arr[i] = i
+    }
+    memoryAllocations.push(arr)
+  }
+
+  class MockConnection extends Connection {
+    request (params, callback) {
+      const stream = intoStream(JSON.stringify({ hello: 'world' }))
+      stream.statusCode = 200
+      stream.headers = {
+        'content-type': 'application/json;utf=8',
+        'content-length': contentLength,
+        connection: 'keep-alive',
+        date: new Date().toISOString()
+      }
+      stream.on('close', () => t.pass('Stream destroyed'))
+      process.nextTick(callback, null, stream)
+      return { abort () { } }
+    }
+  }
+
+  const client = new Client({ node: 'http://localhost:9200', Connection: MockConnection, memoryCircuitBreaker })
+  client.info((err, result) => {
+    t.ok(err instanceof errors.RequestAbortedError)
+    t.equal(err.message, `The content length (${contentLength}) is bigger than the maximum allowed heap memory limit.`)
+    t.equal(result.meta.attempts, 0)
+  })
+})
+
 test('Prototype poisoning protection enabled by default', t => {
   t.plan(1)
 


lib/Transport.js
```diff
@@ -261,6 +265,13 @@ class Transport {
           return onConnectionError(
             new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed string (${MAX_STRING_LENGTH})`, result)
           )
+        } else if (this.memoryCircuitBreakerEnabled && (contentLength + heapUsed > HEAP_SIZE_LIMIT * this.responseMaxHeapPercentage)) {
+          // Memory circuit breaker enabled and the response payload is too large to fit into available heap memory.
+          // Abort this response to avoid OOM crash of dashboards.
+          response.destroy()
+          return onConnectionError(
+            new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed heap memory limit`, result)
```
mihirsoni, line 273: Neat: would be great to add the settings they can tune this setting. i.e `You could tune this via `config.name`

zuochengding: Yes, I will have a follow up PR to add setting in OSD config (opensearch_dashbaords.yml). This PR is only for change in client repo.

zuochengding: Will upgrade the Error message as well to add more instruction guidance

zuochengding:
> Since this is a new feature, would you be able to bump the package.json to be `1.1.0`, for example: [68eea00](https://github.com/opensearch-project/opensearch-js/commit/68eea00b078710c876bd4c65f7b817c8d418a25f)?

Sure, will update, is there any other process to be aware of apart from the version bump?

lib/Transport.js
```diff
@@ -247,10 +250,11 @@ class Transport {
 
       const contentEncoding = (result.headers['content-encoding'] || '').toLowerCase()
       const isCompressed = contentEncoding.indexOf('gzip') > -1 || contentEncoding.indexOf('deflate') > -1
-
       /* istanbul ignore else */
       if (result.headers['content-length'] !== undefined) {
         const contentLength = Number(result.headers['content-length'])
+        const heapUsed = process.memoryUsage().heapUsed
+        // nodeJS data type limit check
```
zengyan-amazon, line 257: move this down to the `else if` bock so that we don't need to get memory usage when circuit break is disabled

package.json
```diff
@@ -11,7 +11,7 @@
     "./": "./"
   },
   "homepage": "https://www.opensearch.org/",
-  "version": "1.0.2",
+  "version": "1.1.0",
```
zengyan-amazon, line 14: is this planned ?

index.d.ts
```diff
@@ -100,6 +100,8 @@ interface ClientOptions {
   maxRetries?: number;
   requestTimeout?: number;
   pingTimeout?: number;
+  responseMaxHeapPercentage?: number | Number;
```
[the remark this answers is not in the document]
zuochengding: will do

lib/Transport.js
zuochengding: make sense, will update. Since i used this variable in else if check expression. I might need create a nested if else statements here, e.g.:
```
else if (this.memoryCircuitBreakerEnabled) {
      const heapUsed = process.memoryUsage().heapUsed
      if (contentLength + heapUsed > ....) {...}
}
```

package.json
zuochengding: I am not aware of the planning yet. following up on @kavilla 's comment.

zuochengding: I am not aware of the planning yet. following up on @kavilla 's comment.

lib/Transport.js
zuochengding: make sense, will update. Since i used this variable in else if check expression. I might need create a nested if else statements here, e.g.:
```
else if (this.memoryCircuitBreakerEnabled) {
      const heapUsed = process.memoryUsage().heapUsed
      if (contentLength + heapUsed > ....) {...}
}
```

zengyan-amazon: sounds good, and please create functions to improve the readability when necessary

zengyan-amazon (approved):

ananzh (approved):

lib/Transport.js
```diff
@@ -261,6 +264,12 @@ class Transport {
           return onConnectionError(
             new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed string (${MAX_STRING_LENGTH})`, result)
           )
+        } else if (shouldApplyCircuitBreaker(contentLength)) {
```
kavilla, line 267: Could this be `this.memoryCircuitBreakerEnabled && shouldApplyCircuitBreaker(contentLength)`?

I would like to avoid `process.memoryUsage()` being called if this is disabled. Based on the [docs](https://nodejs.org/dist/latest-v14.x/docs/api/process.html#process_process_memoryusage):

> The process.memoryUsage() method iterates over each page to gather information about memory usage which might be slow depending on the program memory allocations.

So it seems to potentially expensive operation if used on every request.

index.d.ts
```diff
@@ -100,6 +100,8 @@ interface ClientOptions {
   maxRetries?: number;
   requestTimeout?: number;
   pingTimeout?: number;
+  responseMaxHeapPercentage?: number;
```
kavilla, line 103: This options go together like 🍪 & 🥛,

I think we should consider something like:
```
memoryCircuitBreaker?: {
  enabled: boolean | false;
  maxHeapPercentage: number | 1.0;
}
```

lib/Transport.js
```diff
@@ -81,6 +82,8 @@ class Transport {
     this.opaqueIdPrefix = opts.opaqueIdPrefix
     this[kCompatibleCheck] = 0 // 0 = to be checked, 1 = checking, 2 = checked-ok, 3 checked-notok
     this[kApiVersioning] = process.env.OPENSEARCH_CLIENT_APIVERSIONING === 'true'
+    this.responseMaxHeapPercentage = opts.responseMaxHeapPercentage
```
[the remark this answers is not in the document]
kavilla: There doesn't seem to be something validating this is between 0-1. Seems like I can enable the circuit breaker set it max heap percent to be `75` when I meant for it to be 75% now the function will be:

```
this.memoryCircuitBreakerEnabled && (contentLength + heapUsed > HEAP_SIZE_LIMIT * 75)
```

which doesn't make it useful.

kavilla (changes requested):
Some comments, mostly tied to performance.

Also, it seems like the [doc site](https://opensearch.org/docs/latest/clients/javascript/) [[source code](https://github.com/opensearch-project/documentation-website/blob/main/_clients/javascript.md)]

Seems to mimic the readme section of this repo. I think it will be worth to update some documentation whether it be in the docs repo only under a new section or new section here or in both. Depending on the functionality it might not be clear. Like I probably wouldn't know that valid values are between 0-1. If we could get a doc or an issue in the doc repo that is linked to this PR with the content speaking about how to set the value, expected params, when we are accessing the heap limit, I think that will help provide insight to the community without them going deep into the code.

\m/ Awesome work though \m/

This would be huge benefit to the community even if there is a trade off of performance because we are trying to prevent crashing. In my opinion stability > performance but we should be conscience about the toll on performance if there is a substantial one.

Thank you!

lib/Transport.js
```diff
@@ -51,6 +51,7 @@ const clientVersion = require('../package.json').version
 const userAgent = `opensearch-js/${clientVersion} (${os.platform()} ${os.release()}-${os.arch()}; Node.js ${process.version})`
 const MAX_BUFFER_LENGTH = buffer.constants.MAX_LENGTH
 const MAX_STRING_LENGTH = buffer.constants.MAX_STRING_LENGTH
+const HEAP_SIZE_LIMIT = require('v8').getHeapStatistics().heap_size_limit
```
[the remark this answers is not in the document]
kavilla: Gotcha makes for the v8 usage then. One more clarifying question if I don't set max-old-space-size is it still static like is there a default size?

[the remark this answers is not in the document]
zuochengding: 1. This is limit for Node instead runtime process. This value is static at time of instantiation and configurable through: `NODE_OPTIONS="$NODE_OPTIONS --max-old-space-size=4096"` (MB)
2. If the total memory for OSD (by process) needed exceed this limit, lets say 4GB above, then it will crash as OOM 
`FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`

lib/Transport.js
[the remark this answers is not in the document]
zuochengding: Nice catch here! I agreed that a validator here is necessary.

lib/Transport.js
```diff
@@ -301,6 +310,12 @@ class Transport {
       response.on('end', onEnd)
       response.on('aborted', onAbort)
     }
+    // Helper function to check if memory circuit breaker enabled and the response payload is too large to fit into available heap memory.
+    const shouldApplyCircuitBreaker = (contentLength) => {
+      const heapUsed = process.memoryUsage().heapUsed
```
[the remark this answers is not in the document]
zuochengding: Due to garbage collection can happens at any time. I think we need to the capture the real time memory usage here.

lib/Transport.js
```diff
@@ -261,6 +264,12 @@ class Transport {
           return onConnectionError(
             new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed string (${MAX_STRING_LENGTH})`, result)
           )
+        } else if (shouldApplyCircuitBreaker(contentLength)) {
+          // Abort this response to avoid OOM crash of dashboards.
+          response.destroy()
+          return onConnectionError(
+            new RequestAbortedError(`The content length (${contentLength}) is bigger than the maximum allowed heap memory limit. You could tune this via 'opensearch.responseMaxHeapPercentage' in opensearch_dashboards.yml`, result)
```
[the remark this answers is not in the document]
zuochengding: Okay, make sense

lib/Transport.js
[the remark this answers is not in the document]
zuochengding: Yes, there will be a default value ( 512 MB in latest node version i believe), and it can be different based on the system architecture. For example, the value on 64 bit system can be different from 32bit system.

lib/Transport.js
zuochengding: I may keep the else if statement simple and  do fail fast check first in the helper function like : 
```
shouldApplyCircuitBreaker(contentLength) => {
     if (! this.memoryCircuitBreakerEnabled) return false
     else ....
}
```
Do you agree on this?

kavilla: \m/ Yeah that's cool too!

lib/Transport.js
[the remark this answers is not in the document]
kavilla: True but I'm not sure if this possible but will contentLength ever be bigger than HEAP_SIZE_LIMIT and max heap %? If so I don't see why would need to access the heapUsed since we already know contentLength is already passed the threshold.

For example:

```
const threshold = HEAP_SIZE_LIMIT * this.responseMaxHeapPercentage;
if (contentLength > threshold) return true;
const heapUsed = process.memoryUsage().heapUsed;
return contentLength + heapUsed > threshold;
```

The first two lines should be negligible in runtime compared to when `heapUsed` is accessed. But if some application uses this circuit breaker and hits OpenSearch a lot that returns a lot of results at least we can try to capture the events where the contentLength was just way too big before we access the heap used. WDYT?

lib/Transport.js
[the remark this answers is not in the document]
kavilla: Cool. I wonder for any documentation we could provide that insight, not super explicit, or do you think that would be redundant?

lib/Transport.js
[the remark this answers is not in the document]
zuochengding: Thanks, I see your point now.
I agree that this is nice to have but in quite extreme case. If you see the overall `contentLength` `if else`check, we start with `MAX_BUFFER_LENGTH` and `MAX_STRING_LENGTH` check (The constraint varies based on nodeJs version, lets say node 14, then `MAX_STRING_LENGTH == 512 MB`).  So the first and second checks will filter out most cases when `contentLength` is too large.  When we enter the new added `else if` block, the most affected factor in this scenario will be the `current used heap but not garbage collected yet`. That is why I think `process.memoryUsage().heapUsed` should be called in most cases. 
Your suggestion is also quite valid, I am good to update

lib/Transport.js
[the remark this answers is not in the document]
zuochengding: I have these noted in my design doc. Maybe I can note more down in the PR description?

index.js
```diff
@@ -123,7 +123,9 @@ class Client extends OpenSearchAPI {
         context: null,
         proxy: null,
         enableMetaHeader: true,
-        disablePrototypePoisoningProtection: false
+        disablePrototypePoisoningProtection: false,
```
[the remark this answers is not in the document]
zuochengding: This is no needed, made some change in new revision. Will resolve this conversation

lib/Transport.js
```diff
@@ -636,6 +651,11 @@ function lowerCaseHeaders (oldHeaders) {
   return newHeaders
 }
 
+function validateMemoryPercentage (percetage) {
```
kavilla, line 654: nit: `percetage` typo

index.d.ts
```diff
@@ -126,6 +126,8 @@ interface ClientOptions {
     password?: string;
   };
   disablePrototypePoisoningProtection?: boolean | 'proto' | 'constructor';
+  responseMaxHeapPercentage?: number;
```
kavilla, line 129: nit: i still think we can make this an object since `responseMaxHeapPercentage` is not used unless `memoryCircuitBreakerEnabled`, is there a specific reason why it can't?

kavilla (commented):
LGTM! One comment about the config but not a blocker.

index.d.ts
zuochengding: will do

lib/Transport.js
```diff
@@ -81,6 +82,9 @@ class Transport {
     this.opaqueIdPrefix = opts.opaqueIdPrefix
     this[kCompatibleCheck] = 0 // 0 = to be checked, 1 = checking, 2 = checked-ok, 3 checked-notok
     this[kApiVersioning] = process.env.OPENSEARCH_CLIENT_APIVERSIONING === 'true'
+    this.memoryCircuitBreaker = opts.memoryCircuitBreaker
+    // this.responseMaxHeapPercentage = validateMemoryPercentage(opts.responseMaxHeapPercentage)
```
kavilla, line 86: This can be deleted now if all works fine!

lib/Transport.js
```diff
@@ -636,6 +653,11 @@ function lowerCaseHeaders (oldHeaders) {
   return newHeaders
 }
 
+function validateMemoryPercentage (percentage) {
+  if (percentage < 0 || percentage > 1) return 1.0
```
kavilla, line 657: Seems like percentage could be nullable right but it has a default value so will it ever be null?

zuochengding: It wont be null since we have the default value assgined.

kavilla:
```
  memoryCircuitBreaker?: {
    enabled: boolean | false;
    maxPercentage?: number | 1.0;
  }
```

I think I led you astray. Your object's properties consist of union types. So `enabled `being `boolean | false` is redundant because we say enabled can only be of type `boolean`. Where as `maxPercentage` which has is optional with the `?` has the redundant `number | 1.0` by it an optional type that is actually `number | undefined` so someone can create a memoryCircuitBreaker   and it won't have default values unless there's a file that you are defining defaults here. But in the end this is editing a JS file so these values can be undefined even if you define a type.

zuochengding: I think I confused the use of `|` in javascript type definition. I thought it will do similar thing as logical operator `value || defaultValue`. I intentionally tested this scenario and passed because In OSD, I have  default value assigned for `memoryCircuitBreaker` object. Thanks for pointing out. will update

kavilla (approved):
```
 #        #####  ####### #     # 
 #       #     #    #    ##   ## 
 #       #          #    # # # # 
 #       #  ####    #    #  #  # 
 #       #     #    #    #     # 
 #       #     #    #    #     # 
 #######  #####     #    #     #                  
```

kavilla:
@zuochengding  do you have write permissions to merge? Also, do you have a follow up to update some documentation or create an issue in the docs repo?

zuochengding:
> @zuochengding do you have write permissions to merge? Also, do you have a follow up to update some documentation or create an issue in the docs repo?

Create a issue in website repo to track the follow up : https://github.com/opensearch-project/documentation-website/issues/470

kavilla:
> > @zuochengding do you have write permissions to merge? Also, do you have a follow up to update some documentation or create an issue in the docs repo?
> 
> Looks like i don't have permission to merge. Can you guys help the merge and cut a `1.1.0` branch? Created a issue in website repo to track the follow up : [opensearch-project/documentation-website#470](https://github.com/opensearch-project/documentation-website/issues/470)

dblock:
@zuochengding Did this introduce https://github.com/opensearch-project/opensearch-js/issues/327? Care to take a look?

PR #210, opened 2022-03-20
[FEATURE] Support for script operations in the update action of the client.helpers.bulk method

### LINKED ISSUE #209: [FEATURE] helpers.bulk support script update
**Is your feature request related to a problem?**
We are currently using AWS OpenSearch for a project in our business.
And, to our delight, a lot of data is now stored.
However, recently I am getting 429 errors returned when creating, updating, or deleting documents using the client.bulk method.
We have tried to use client.helpers.bulk to prevent this as much as possible.
However, we found that updating documents is only supported with a doc update query.

**What solution would you like?**
I would like to use client.helpers.bulk for script update as well.

**What alternatives have you considered?**
We have already tried script update with client.bulk, but
I would like to create, update (doc and script) and delete documents flexibly with client.helper.bulk.
In particular, we would like to make safe API requests using the concurrency and flushBytes options.

huuyafwww:
Signed-off-by: huuyafwww [huuya1234fwww@gmail.com](mailto:huuya1234fwww@gmail.com)

### Description
This is a modification to the client.helpers.bulk method to support script operations in the update action.
 
### Issues Resolved
#209
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/lib/Helpers.d.ts
+++ b/lib/Helpers.d.ts
@@ -82,6 +82,18 @@ interface UpdateActionOperation {
   }
 }
 
+interface UpdateActionDocOperation {
+  doc: {
+    [key: string]: any
+  }
+}
+
+interface UpdateActionScriptOperation {
+  script: {
+    [key: string]: any
+  }
+}
+
 interface DeleteAction {
   delete: {
     _index: string
@@ -89,7 +101,10 @@ interface DeleteAction {
   }
 }
 
-type UpdateAction = [UpdateActionOperation, Record<string, any>]
+type UpdateAction = [
+  UpdateActionOperation,
+  UpdateActionDocOperation | UpdateActionScriptOperation
+];
 type Action = IndexAction | CreateAction | UpdateAction | DeleteAction
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
 
--- a/lib/Helpers.js
+++ b/lib/Helpers.js
@@ -524,9 +524,7 @@ class Helpers {
           bulkBody.push(actionBody, payloadBody)
         } else if (operation === 'update') {
           actionBody = serializer.serialize(action[0])
-          payloadBody = typeof chunk === 'string'
-            ? `{"doc":${chunk}}`
-            : serializer.serialize({ doc: chunk, ...action[1] })
+          payloadBody = serializer.serialize(action[1] || chunk)
           chunkBytes += Buffer.byteLength(actionBody) + Buffer.byteLength(payloadBody)
           bulkBody.push(actionBody, payloadBody)
         } else if (operation === 'delete') {
--- a/test/types/helpers.test-d.ts
+++ b/test/types/helpers.test-d.ts
@@ -36,7 +36,8 @@ import {
   BulkHelperOptions,
   ScrollSearchResponse,
   OnDropDocument,
-  MsearchHelper
+  MsearchHelper,
+  UpdateActionDocOperation
 } from '../../lib/Helpers'
 import { ApiResponse, ApiError, Context } from '../../lib/Transport'
 
@@ -108,7 +109,7 @@ expectError(
   const options: BulkHelperOptions<Record<string, any>> = {
     datasource: [],
     onDocument(doc: Record<string, any>) {
-      return [{ update: { _index: 'test' } }, doc]
+      return [{ update: { _index: 'test' } }, doc as UpdateActionDocOperation]
     }
   }
   expectAssignable<BulkHelperOptions<Record<string, any>>>(options)
--- a/test/unit/helpers/bulk.test.js
+++ b/test/unit/helpers/bulk.test.js
@@ -848,12 +848,14 @@ test('bulk update', t => {
       flushBytes: 1,
       concurrency: 1,
       onDocument (doc) {
+        const currentId = id++
         return [{
           update: {
             _index: 'test',
-            _id: id++
+            _id: currentId
           }
         }, {
+          doc: dataset[currentId],
           doc_as_upsert: true
         }]
       },
@@ -896,12 +898,13 @@ test('bulk update', t => {
       flushBytes: 1,
       concurrency: 1,
       onDocument (doc) {
+        const currentId = id++
         return [{
           update: {
             _index: 'test',
-            _id: id++
+            _id: currentId
           }
-        }]
+        }, { doc: dataset[currentId] }]
       },
       onDrop (doc) {
         t.fail('This should never be called')
@@ -942,12 +945,14 @@ test('bulk update', t => {
       flushBytes: 1,
       concurrency: 1,
       onDocument (doc) {
+        const currentId = id++
         return [{
           update: {
             _index: 'test',
-            _id: id++
+            _id: currentId
           }
         }, {
+          doc: dataset[currentId],
           doc_as_upsert: true
         }]
       },


huuyafwww:
Here is the bulkBody after and before modification.
This modification makes the bulkBody I wanted!

sample code here:

```js
const body = [[
  { update: { _id: '1', _index: 'sample-index'} },
  { script: {
      source: "ctx._source.body = params.body;",
      lang: 'painless',
      params: {
        body: "sample body",
      }
    },
    scripted_upsert: true,
  },
]];

await client.helpers.bulk({
  datasource: body,
  onDocument (doc) {
    return doc;
  }
});
```

before bulkBody
```json
{"update":{ "_id": "1", "_index": "sample-index"}},
{"doc":[{"update":{ "_id": "1", "_index": "sample-index"}},{"script":{"source":"ctx._source.body = params.body;","lang":"painless","params":{"body":"sample body"}},"scripted_upsert":true} 
```

after bulkBody
```json
{"update":{ "_id": "1", "_index": "sample-index"}},
{"script":{"source":"ctx._source.body = params.body;","lang":"painless","params":{"body":"sample body"}},"scripted_upsert":true}
```

kavilla:
Hello @huuyafwww, thanks for opening this! Since this changes a request, I would consider this a major version change which could impact other consumers of the package. To build this in a way that makes it possible to get your desired behavior by making it configurable with the default path being the current request.

huuyafwww:
@kavilla 
Thanks for the check.

> Since this changes a request, I would consider this a major version change which could impact other consumers of the package. 

Indeed it is.
I agree with you.

> To build this in a way that makes it possible to get your desired behavior by making it configurable with the default path being the current request.

Ok.
Am I correct in assuming that we build a "doc" or "script" depending on the contents of the `chunk` ?

huuyafwww:
@kavilla 

> Am I correct in assuming that we build a "doc" or "script" depending on the contents of the chunk or action?

This is fine, but if you can determine that the operation is a "doc" or "update" operation by the code below, for example, you know that the developer has entered "doc" or "update" as the return value of onDocument in the first place.

```js
const document = Array.isArray(action)
  ? Object.keys(action[1])[0]
  : Object.keys(action)[0]
```

So we assigned `action[1]` as the completion of `Object.keys(action[1])[0]` and `chunk` as the completion of `Object.keys(action)[0]`.

```js
payloadBody = serializer.serialize(action[1] || chunk)
```

What do you think about my idea?

huuyafwww:
Hi, @kavilla!
What do you think of my opinion?
Thanks for taking the time to respond.

kavilla:
Hello @huuyafwww, will take a look next week if that is okay! Sorry about the delay.

huuyafwww:
Hello @kavilla, Thanks again for your thoughtful replies.
Understood!

kavilla:
Sorry for the delay, I was incapacitated for the past two weeks! I see what you are getting at and makes sense for me. But since having forked this repo I will need to derive the same level of insight as you. So what you are saying is that by the time the update operation is hit there was essentially no reason why it was checking if it was a doc or not since the value should be correct and that we were previously spreading the completion into the payload body which was the root case of script operations not being supported in update operations. Because basically it was getting incorrectly applied everytime?

huuyafwww:
> Sorry for the delay, I was incapacitated for the past two weeks!

Don't worry about it.
Thank you for confirming my request in spite of all that.

> But since having forked this repo I will need to derive the same level of insight as you.

Roger that!

> So what you are saying is that by the time the update operation is hit there was essentially no reason why it was checking if it was a doc or not since the value should be correct and that we were previously spreading the completion into the payload body which was the root case of script operations not being supported in update operations. Because basically it was getting incorrectly applied everytime?

You're right!
There are doc and script for api batch update operations, but only doc was supported by this helpers.bulk in the first place.
The implementation was based on the doc premise and was designed to be complementary, so there was no way to request an update operation in the script.
Therefore, I am modifying the relevant section in this pull request to also support script updates in this helpers.bulk.

Also, as you can see, there was no need for a conventional complementary implementation in the first place.
This is because if this library can determine whether the update operation the developer is trying to perform is doc or update, there is no need for completion in the first place.

In addition, to explain, being able to determine whether the payload body of an update operation is doc or update is no different from receiving that information directly from the return value of onDocument.

Therefore, since it is no longer necessary to supplement the payload body as before, we have modified it in this way because we believe that the best implementation is to use the return value given by the developer using this library as-is as the payload body.

huuyafwww:
@kavilla
Hi kavilla.
How about what I am proposing in this pull request?

I have already patched this pull request change to the business system and am running it in the production environment.
If possible, I would like to see it taken in.

huuyafwww:
Since this pull request was created, the contents of the main branch had changed and was no longer working with the latest version, so the main branch was imported and modified.

dblock:
Thanks. Needs tests for `UpdateActionScriptOperation` and a changelog entry.

huuyafwww:
@dblock 

> Thanks. Needs tests for `UpdateActionScriptOperation` and a changelog entry.

Sorry for the delay in responding.
I tested UpdateActionScriptOperation and added an entry to the changelog.
I also added a bulk update test for script operations.

kavilla (approved):
lgtm. Any issues?

huuyafwww:
@kavilla 

> lgtm. Any issues?

I can no longer bulk update doc operations in the conventional way,
It involves some disruptive changes.

wbeckler:
@nhtruong would you be willing to take a look, especially with respect to backwards compatibility?

lib/Helpers.js
```diff
@@ -539,10 +539,7 @@ class Helpers {
           bulkBody.push(actionBody, payloadBody);
         } else if (operation === 'update') {
           actionBody = serializer.serialize(action);
-          payloadBody =
-            typeof chunk === 'string'
-              ? `{"doc":${chunk}}`
-              : serializer.serialize({ doc: chunk, ...payload });
```
nhtruong, line 545: Is this the reason why this breaks the bulk operation for regular use of this helper method?

(Also sorry the comment mentioning me got lost in the mail)

lib/Helpers.d.ts
```diff
@@ -111,7 +123,10 @@ interface DeleteAction {
 
 type CreateAction = CreateActionOperation | [CreateActionOperation, unknown];
 type IndexAction = IndexActionOperation | [IndexActionOperation, unknown];
-type UpdateAction = [UpdateActionOperation, Record<string, any>];
```
nhtruong, line 114: Does this mean the old method already supports `script` since it would accept anything with `Record<string,any>`?

any-yamauchi: @nhtruong 
The old method accepted script input with Record<string,any> as the type, but since script was not supported in the first place, unintended queries were generated.

huuyafwww: @nhtruong 
The old method accepted script input with Record<string,any> as the type,
but since script was not supported in the first place, unintended queries were generated.

lib/Helpers.js
huuyafwww: @nhtruong 
Yes. The old method only supported doc, so we modified it to support script as well, but that will not work with the previous method of input.

nhtruong: We should avoid breaking changes to existing users. Unless there's no other way around this?

lib/Helpers.d.ts
nhtruong: Gotcha. I thought the users couldn't do script before from the title of the PR.

dblock:
Looks like we failed to merge this after all. @nhtruong could you please take a look?

[Catch All Triage - Attendees [1](https://github.com/dblock/), [2](https://github.com/krisfreedain), [3](https://github.com/andrross), [4](https://github.com/gkamat), [5](https://github.com/getsaurabh02)]

nhtruong:
Will add this to the 3.0.0 release and this introduces a breaking change.

PR #228, opened 2022-04-29
Add support to test against multiple versions of OpenSearch

### LINKED ISSUE #214: Test and publish compatibility matrix with various OpenSearch versions
Add support to run integration tests across various versions of OpenSearch and publish a compatibility matrix for the client version vs OpenSearch version.

PR Reference: https://github.com/opensearch-project/opensearch-java/pull/122
Compatibility: https://github.com/opensearch-project/opensearch-java/blob/main/COMPATIBILITY.md

VachaShah:
### Description
Adding support to run tests against multiple versions of OpenSearch in the CI. I separated out the `compatibility.yml` since the matrix was becoming huge in the `integration.yml` with multiple node js versions.
 
### Issues Resolved
#214 
 
### Check List
  - [X] All tests pass

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -8,23 +8,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        secured: ["true", "false"]
         entry:
-          - { version: '1.1.0', secured: 'true' }
-          - { version: '1.1.0', secured: 'false' }
-          - { version: '1.2.0', secured: 'true' }
-          - { version: '1.2.0', secured: 'false' }
-          - { version: '1.2.1', secured: 'true' }
-          - { version: '1.2.1', secured: 'false' }
-          - { version: '1.2.2', secured: 'true' }
-          - { version: '1.2.2', secured: 'false' }
-          - { version: '1.2.3', secured: 'true' }
-          - { version: '1.2.3', secured: 'false' }
-          - { version: '1.2.4', secured: 'true' }
-          - { version: '1.2.4', secured: 'false' }
-          - { version: '1.3.0', secured: 'true' }
-          - { version: '1.3.0', secured: 'false' }
-          - { version: '1.3.1', secured: 'true' }
-          - { version: '1.3.1', secured: 'false' }
+          - { opensearch_version: 1.1.0 }
+          - { opensearch_version: 1.2.0 }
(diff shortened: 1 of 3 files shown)


dblock (approved):

dblock:
> Hello, can i already consider this compatibility matrix as truth? We have a cluster with `opensearch 1.2` on aws, and we are developing our client based on `@opensearch-project/opensearch@1.1`, we saw that `elasticsearch client version` should match the `cluster version`, but for opensearch we could use `opensearch client 1.1 `for `1.2` cluster right?

Correct. Generally, we're following semver and ensuring there's an easy upgrade path that does not require upgrading in lockstep. Please do check out https://github.com/opensearch-project/opensearch-clients/issues/17 and add your comments!

.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,66 @@
...
+on: [push, pull_request]
+
+jobs:
+  integration-test-compatibility:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        entry:
+          - { version: '1.1.0', secured: 'true' }
```
saratvemulapalli, line 12: May be Im totally off, I see we are replicating few pieces of code in all client repositories. 
Is there a way we can define these in a single plane (May be opensearch-clients) and consume it in all our client repos?

saratvemulapalli (approved):

.github/workflows/compatibility.yml
VachaShah: @dblock Updated the workflow matrix.

VachaShah: @saratvemulapalli Do you mean like having a template in `opensearch-clients`?

saratvemulapalli: I dont know what a template is :/. 
I was looking for a single place where common code can live. Lets take compatibility as an example:
- Defining this in a central place will make it easier when we add new version support.
- Enforces our CIs to test against new versions.
- Keeps all clients in sync.

VachaShah: Oh got it! Yeah we are trying to keep it as consistent as possible across the client repos but I will keep this in mind if we can place common code. My current thought is having a single place might be difficult since the clients can be on different versions with different compatibility, their CIs are written differently etc. But will explore this once the breaking changes are done.

PR #266, opened 2022-07-21
Add missing memory circuit breaker options

robdasilva:
### Description
Add missing `memoryCircuitBreaker` option in `Transport.d.ts`.

The discussion in https://github.com/opensearch-project/documentation-website/issues/823 revealed that this option appeared to be missing: https://github.com/opensearch-project/documentation-website/issues/823#issuecomment-1191962709

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -45,6 +45,7 @@ import Transport, {
   RequestBody,
   RequestNDBody,
   Context,
+  MemoryCircuitBreakerOptions,
 } from './lib/Transport';
 import { URL } from 'url';
 import Connection, { AgentOptions, agentFn } from './lib/Connection';
@@ -137,10 +138,7 @@ interface ClientOptions {
     password?: string;
   };
   disablePrototypePoisoningProtection?: boolean | 'proto' | 'constructor';
-  memoryCircuitBreaker?: {
-    enabled: boolean;
-    maxPercentage: number;
-  };
+  memoryCircuitBreaker?: MemoryCircuitBreakerOptions
 }
 
 declare class Client {
--- a/lib/Transport.d.ts
+++ b/lib/Transport.d.ts
@@ -29,10 +29,10 @@
  */
 
 import { Readable as ReadableStream } from 'stream';
-import { ConnectionPool, CloudConnectionPool } from './pool';
 import Connection from './Connection';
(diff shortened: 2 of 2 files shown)


kavilla (approved):
Thank you so much for adding this

ananzh (approved):
nice. thanks a lot

dblock:
Looks like a test failed?

We need tests for this, don't we?

robdasilva:
@dblock The integration test appears to have failed due to no tests being found. So there seem to exist no tests yet for the client helpers. As this PR merely adds an option to the type declarations (...d.ts) that already exists in code, the changes in this PR should ink y understanding not have the potential to break existing behavior. Do you have any concerns?

dblock:
Looks like tests are failing on main, I opened https://github.com/opensearch-project/opensearch-js/issues/268 (someone should please take care of this).

We could merge this PR as is, it doesn't break anything existing, but I would expect any new code (this new option added) to be exercised in tests so that we don't break it in the future.

robdasilva:
> I would expect any new code (this new option added) to be exercised in tests so that we don't break it in the future.

@dblock Absolutely, I agree. What I mentioned above though, is that this PR does not add the option. It was there before already. This PR only adds it to the type declarations.

How would the types be tested to contain all options actually available in the code?

dblock:
I see. I think we should rewrite this client in TypeScript :)

How do we know if other options are not missing?

dblock:
I merged this, thanks.

robdasilva:
> How do we know if other options are not missing?

@dblock Rewriting in TypeScript would be one option. But other than that it's a tricky question. I'd be curios to know myself.

The existing TypeScript types could actually be used in the JS code [using JSDoc comments](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html), but that would only help with code completion. Not necessarily with ensuring there is no mismatch between the code and manually written TS declarations.

Using JSDoc to define the parameters and generating TS types from that might be an option. But I have little experience with that tbh.

dblock:
I opened https://github.com/opensearch-project/opensearch-js/issues/269 to gather input on whether anyone feels strongly about rewriting this client in TypeScript, or wants to take a stab at it!

Issue #187, opened 2021-10-21
[FEATURE] Add IAM authentication option


aetter:
If people want to connect to an Amazon OpenSearch Service cluster, they have to figure out how to use this client alongside some third-party signing library and/or the AWS SDK, which is non-trivial due to how we (by design) conceal the underlying HTTP requests that the client makes. We should offer IAM signing as an option, just like the OpenSearch CLI does.

**Is your feature request related to a problem?**

If you want to use this client with an Amazon OpenSearch Service cluster that has IAM authentication rather than basic authentication, good luck.

**What solution would you like?**

When initializing the client, accept an iam_auth option in place of the current basic auth option. If basic, accept username and password. If IAM, accept options for access key, secret key, session token, region, and service.

**What alternatives have you considered?**

An additional JS signing library for use on top of the client, such as https://github.com/compwright/aws-elasticsearch-connector. But this is third-party, might not use the latest version of the JS SDK, and given that it would only work with this client, it seems better and easier to just add it to the client and maintain it ourselves.

**Do you have any additional context?**

https://github.com/opensearch-project/OpenSearch/issues/1400

van-vothanh:
have you tried this approach https://docs.aws.amazon.com/opensearch-service/latest/developerguide/request-signing.html#request-signing-node?

Ghazgkull:
Until OpenSearch adds authentication support for hosted offerings like AWS OpenSearch, is there a workaround available for users of those hosted offerings? I'm confused about the current situation... how is anyone using the OpenSearch JS client today, without support for authentication?

rawpixel-vincent:
until the role authentication is supported, we're stuck into the latest supported version of elasticsearch with the latest elasticsearch node.js client compatible with a library that wrap that client to sign the requests.
we're looking forward to get this support so we can upgrade to opensearch 1.0

rawpixel-vincent:
I think it should take around 1 day to get it done and hopefully around 2 weeks to fix/review/release it? that would be amazing and aws customers could put all the pain behind and enjoy an amazing truly open source service, what do you think?

ChristopheBougere:
This is an important feature for us as well, I hope this will be implemented soon 🙏

alexsasharegan:
[...]

PR #279, opened 2022-08-25
Add AwsSigV4 signing functionality

### LINKED ISSUE #252: [FEATURE] Add support for AWS Sigv4 request signer
Add support for signed requests via AWS sigv4.

harshavamsi:
Signed-off-by: Harsha Vamsi Kalluri <harshavamsi096@gmail.com>

### Description
Adds the ability to sign requests natively using awsSigV4. An example is given in the README on how to sign every request.

### Issues Resolved
Fixes #252 


kavilla:
did we want to bump the minor version to reflect the new feature? https://github.com/opensearch-project/opensearch-js/blob/main/package.json#L14

lib/AwsV4Signer.d.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Credentials } from '@aws-sdk/types';
+import Connection from './Connection';
+import * as http from 'http';
+
+interface AwsV4SignerOptions {
+  credentials: Credentials;
+  region: string;
+}
+
+export interface AwsV4SignerResponse {
```
kavilla, line 40: nit: is `Response` standard? If it is I'm cool with it but for me feels like Response for something that is happening pre-flight feels off

kavilla (approved):
This is so awesome! Thank you so much.

Might need to ping docs team to prepare themselves and then release this.

kavilla (commented):
Comment for nits and questions.

README.md
```diff
@@ -152,6 +152,93 @@ async function search() {
...
+  var settings = {
+    settings: {
+      index: {
+        number_of_shards: 4,
+        number_of_replicas: 3,
+      },
+    },
+  };
+
+  var response = await client.indices.create({
```
dblock, line 187: This is identical to the samples above? Maybe we should break up the examples to specific sections (indexing, searching, AWS Sigv4 Signing...?

lib/AwsV4Signer.d.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
```
dblock, line 12: New code does not need license to Elastic, remove. Only SPDX header remains.

dblock (changes requested):
This is a great start! Haven't looked at the detailed implementation yet. Consider https://github.com/opensearch-project/opensearch-js/issues/187#issuecomment-1221191959 and other comments in that issue.

My biggest ask is to refactor this in some kind of optional, namespaced module. This is vendor-specific code and thus I don't think it should be included by default in everybody's client usage. WDYT?

harshavamsi:
> did we want to bump the minor version to reflect the new feature? https://github.com/opensearch-project/opensearch-js/blob/main/package.json#L14

My assumption was that this would go as a 2.x release instead of a 2.0.x? I could be wrong.

Does it make sense to bump version here of a separate commit before the release?

lib/AwsV4Signer.d.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Licensed to Elasticsearch B.V. under one or more contributor
```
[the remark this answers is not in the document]
harshavamsi: Updated.

index.d.ts
```diff
@@ -5050,4 +5051,5 @@ export {
   ClientOptions,
   NodeOptions,
   ClientExtendsCallbackOptions,
+  AwsV4Signer
```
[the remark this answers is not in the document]
harshavamsi: Fixed.

lib/AwsV4Signer.js
```diff
@@ -0,0 +1,62 @@
...
+function AwsV4Signer(opts) {
+  if (opts && (!opts.region || opts.region === null || opts.region === '')) {
+    throw new AwsV4SignerError('Region cannot be empty');
+  }
+  if (opts && (!opts.credentials || opts.credentials === null || opts.credentials === '')) {
+    throw new AwsV4SignerError('Credentials cannot be empty');
+  }
+
+  function buildSignedRequestObject(request = {}) {
+    request.service = 'es';
```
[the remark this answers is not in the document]
harshavamsi: Yes, we still use `es` as the service name.

README.md
harshavamsi: The reason why it lives in a separate code block is just because of the difference in the imports and the way the client is initialized. Because the signer needs credentials, it's an async call. Felt like maybe the two should be written separately.

harshavamsi:
> This is a great start! Haven't looked at the detailed implementation yet. Consider [#187 (comment)](https://github.com/opensearch-project/opensearch-js/issues/187#issuecomment-1221191959) and other comments in that issue.
> 
> My biggest ask is to refactor this in some kind of optional, namespaced module. This is vendor-specific code and thus I don't think it should be included by default in everybody's client usage. WDYT?

I agree with you. I've separated out the signer from the main injection point and users will have to require it explicitly to use it.

dblock:
> > did we want to bump the minor version to reflect the new feature? https://github.com/opensearch-project/opensearch-js/blob/main/package.json#L14
> 
> My assumption was that this would go as a 2.x release instead of a 2.0.x? I could be wrong.

Yes, if there are no breaking changes and only new features, this is 2.1 per semver. After release we should have incremented the patch version to 2.0.1. So please bump the version in package.json to 2.1.0 as part of this PR?

> Does it make sense to bump version here or a separate commit before the release?

Here.

README.md
```diff
@@ -152,6 +152,94 @@ async function search() {
 search().catch(console.log);
 ```
 
+### With AWS SigV4 signing
+```javascript
+const endpoint = ""; // OpenSearch domain URL e.g. https://search-xxx.region.es.amazonaws.com
+const { Client } = require('@opensearch-project/opensearch');
+const AwsV4Signer = require('@opensearch-project/opensearch/AwsV4Signer');
```
dblock, line 159: I think I'd want it to match above and be namespaced under `aws`:

```javascript
const { AwsSigV4Signer } = require ('@opensearch-project/opensearch/aws')
```

lib/aws/AwsV4Signer.d.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { Credentials } from '@aws-sdk/types';
+import Connection from '../Connection';
+import * as http from 'http';
+
+interface AwsV4SignerOptions {
```
dblock, line 16: So technically this is "AwsSigv4". I think we should align with other implementations such as `Aws::Sigv4::Signer` in Java and include "sig" and lowercase the "v", so `AwsSigv4Signer...`?

dblock (changes requested):
Excellent. I have some asks re:naming of classes/namespaces. Let's get that right since these are hard to change later?

README.md
harshavamsi: This sounds right. I fixed the namespace to allow this kind of import.

lib/aws/AwsV4Signer.d.ts
harshavamsi: Yeah, let's have them be consistent. I renamed everything.

harshavamsi:
> Excellent. I have some asks re:naming of classes/namespaces. Let's get that right since these are hard to change later?
> 
> Also license checker is complaining, I didn't look at the errors, needs to be 🍏.



> Excellent. I have some asks re:naming of classes/namespaces. Let's get that right since these are hard to change later?
> 
> Also license checker is complaining, I didn't look at the errors, needs to be 🍏.

@dblock looks like `@aws-sdk/credential-provider-node` has `tslib` dependency. This is the error that the license checker throws: `"tslib@1.14.1" is licensed under "0BSD" which is not permitted by the --onlyAllow flag`

Not sure what the course of action here is.

README.md
```diff
@@ -152,6 +153,94 @@ async function search() {
...
+  var response = await client.indices.create({
+    index: index_name,
+    body: settings,
+  });
+
+  console.log("Creating index:");
+  console.log(response.body);
+
+  // Add a document to the index.
+  var document = {
```
VachaShah, line 198: All of the examples here look very similar to the already existing ones. I see that initializing the client is different, can we provide may be just one example here so we don't duplicate all the test data?

Also, consider moving this to a USER_GUIDE since the number of examples is increasing, might make the README very long.

dblock:
> @dblock looks like `@aws-sdk/credential-provider-node` has `tslib` dependency. This is the error that the license checker throws: `"tslib@1.14.1" is licensed under "0BSD" which is not permitted by the --onlyAllow flag`
> 
> Not sure what the course of action here is.

Add it to the [list of approved licenses in package.json](https://github.com/opensearch-project/opensearch-js/blob/c9a073441a9f06f8a1c59bdeebf2eece98856a2d/package.json#L38) in this PR, while I go check with someone paid to say whether this is A-OK or not.

README.md
dblock: +1, but I also think we can break this up in this PR or do it in another PR, so no hard ask from me. FYI I do like what they have done in the .NET client: https://github.com/opensearch-project/opensearch-net/blob/main/USER_GUIDE.md

package.json
```diff
@@ -77,6 +88,8 @@
     "xmlbuilder2": "^2.4.1"
   },
   "dependencies": {
+    "@aws-sdk/credential-provider-node": "^3.154.0",
```
dblock, line 91: So does this become a hard dependency for this package? Meaning if I am not doing AWS sigv4, am I still dragging this module in? Asking again because I'd like not to drag any vendor specific components in by default.

README.md
harshavamsi: @VachaShah @dblock I've split the code snippets into a new USER_GUIDE. wdyt?

harshavamsi:
> > @dblock looks like `@aws-sdk/credential-provider-node` has `tslib` dependency. This is the error that the license checker throws: `"tslib@1.14.1" is licensed under "0BSD" which is not permitted by the --onlyAllow flag`
> > Not sure what the course of action here is.
> 
> Add it to the [list of approved licenses in package.json](https://github.com/opensearch-project/opensearch-js/blob/c9a073441a9f06f8a1c59bdeebf2eece98856a2d/package.json#L38) in this PR, while I go check with someone paid to say whether this is A-OK or not.

I just realized that we do not need that package. A user should just get the credentials from his machine and pass them in as arguments. We don't need to deal with that license now.

package.json
harshavamsi: Removed this dependency like I mentioned below. We are still depending on the aws4 library which we cannot do away(unless we want to re-write all of that implementation) if we want to support signing natively in the client.

USER_GUIDE.md
```diff
@@ -0,0 +1,150 @@
...
+  ssl: {
+    ca: fs.readFileSync(ca_certs_path),
+    // You can turn off certificate verification (rejectUnauthorized: false) if you're using self-signed certificates with a hostname mismatch.
+    // cert: fs.readFileSync(client_cert_path),
+    // key: fs.readFileSync(client_key_path)
+  },
+});
+```
+
+## Initializing a Client with AWS SigV4 Signing
```
dblock, line 40: Nit: To authenticate with the [Amazon OpenSearch Service](https://aws.amazon.com/opensearch-service/) use `AwsSigv4Signer`.

USER_GUIDE.md
```diff
@@ -0,0 +1,150 @@
...
+      credentials: credentials,
+      region: "us-west-2",
+    }),
+    node: endpoint,
+  });
+  return client;
+}
+```
+
+## Creating an Index
```
dblock, line 61: Pick one or the other in the headings? 

a) "Create an Index", "Add a Document", "Delete ..." 
b) "Creating an Index", "Adding a Document", "Deleting ..."

dblock (changes requested):
I still have the question on whether we are now requiring vendor-specific aws4 for all scenarios and whether there's a way to avoid that?

 +Some nits.

dblock:
@harshavamsi Looks like you'll also need to rebase CHANGELOG. I think this is almost done!

USER_GUIDE.md
```diff
@@ -0,0 +1,150 @@
...
+```javascript
+const endpoint = ""; // OpenSearch domain URL e.g. https://search-xxx.region.es.amazonaws.com
+const { Client } = require('@opensearch-project/opensearch');
+const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws');
+const { defaultProvider } = require("@aws-sdk/credential-provider-node");
+
+async function getClient() {
+  const credentials = await defaultProvider()();
+  var client = new Client({
+    ...AwsV4Signer({
```
graemecode, line 51: is this meant to be `AwsSigv4Signer` ?

graemecode, line 51:
```suggestion
    ...AwsSigv4Signer({
```

harshavamsi: Missed this, thanks!

harshavamsi:
@dblock all checks are passing, this should be good to go.

rawpixel-vincent:
I've made a port of aws-opensearch-connector inspired by this PR for the docs.

It surely requires more works (maybe reuse the class and test to sign request from this PR) and needs tests.

The intent is to help moving forward the issue with expired credentials

https://github.com/opensearch-project/opensearch-js/pull/289

rawpixel-vincent:
opened https://github.com/harshavamsi/opensearch-js/pull/1

it adds the support for AWS V2 and handle the credentials issue

updated the docs

harshavamsi:
> thank you so much for this @harshavamsi, I haven't tested it just looked at the PR. ~At first glance I'm not sure this take care of renewing the credentials after they expire, or does it ?~ Tested and it does not refresh the credentials after they have expired.
> 
> If a node application register the client once with `getClient()`:
> 
> ```js
> const endpoint = ""; // OpenSearch domain URL e.g. https://search-xxx.region.es.amazonaws.com
> const { Client } = require('@opensearch-project/opensearch');
> const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws');
> const { defaultProvider } = require("@aws-sdk/credential-provider-node");
> 
> async function getClient() {
>   const credentials = await defaultProvider()();
>   var client = new Client({
>     ...AwsSigv4Signer({
>       credentials: credentials,
>       region: "us-west-2",
>     }),
>     node: endpoint,
>   });
>   return client;
> }
> ```
> 
> when the credentials acquired at the time the client was instantiated expires all requests will fail
> 
> Edit: the library we are currently using Extends the Transport class from this library to retrieve the credentials on each request, see: https://github.com/yosefbs/aws-opensearch-connector/blob/master/src/AmazonTransport.js
> 
> Edit2: for the context if you are using the code provided in the aws documentation to sign the request and are using instance role or any other temporary session token credentials based, the client request will stop working once the credentials retrieved when you instantiated the client expires https://docs.aws.amazon.com/opensearch-service/latest/developerguide/request-signing.html#request-signing-node
> 
> Instantiating a client once when the node application start is a common use case. Without any knowledge that this does not take care of refreshing the credentials **this could lead to outage in production environment as testing are often made with either temporary instances or aws user keys** thus it makes it unusable unless you use AWS access keys (not recommended) or reinstantiate the client on every request (probably not a good idea)
> 
> The test to reproduce require a short life session token, I'm not sure if there is a way to use aws credentials and specify an very short lifetime so you could reproduce this case more easily

Excellent catch, thanks for this! I have some questions that I'm going to leave on your PR.

harshavamsi:
@dblock thoughts on https://github.com/harshavamsi/opensearch-js/pull/1#pullrequestreview-1092303311

rawpixel-vincent:
@harshavamsi @dblock opened this as an alternative approach to leave more options to the implementation
https://github.com/harshavamsi/opensearch-js/pull/2

dblock (approved):

dblock:
@harshavamsi @rawpixel-vincent Do we want to merge this as is and consider https://github.com/harshavamsi/opensearch-js/pull/2 on top, or do you want me to wait till you guys discuss that PR?

harshavamsi:
@dblock @rawpixel-vincent I think that we can merge this but I have some concerns. The newer implementation has breaking changes for how this should be used. It is not too dramatic but would be inconsistent if we let this go through and follow up with another set of breaking changes.

On the flip side we haven't completely hashed out the details of that implementation. I can see some potential issues with something like `client.close()` and the connection still being open because of the setTimeout to refresh credentials. I think it would need more tests and will take potentially longer to complete. WDYT?

rawpixel-vincent:
I agree with @harshavamsi it would be best to avoid changing twice the api and wait until we have something that works for everyone. and merging this without taking care of the expired credentials can potentially cause issues for consumer that they won't catch until sending to production.

rawpixel-vincent:
Hi @harshavamsi heads up that there is more coming, will update

rawpixel-vincent:
Hi @harshavamsi @dblock 
opened https://github.com/harshavamsi/opensearch-js/pull/3 that improves the client initialization and improve the sdk credentials expiration/renew handling

dblock:
I'll wait till you both (@harshavamsi and @rawpixel-vincent) are satisfied with the implementation. LMK if I can be of any help!

rawpixel-vincent:
@dblock I've been testing the client with https://github.com/harshavamsi/opensearch-js/pull/3 it works fine and I'm happy with it. 
I'm waiting for @harshavamsi to review and will help to get it ready to merge

rawpixel-vincent:
@dblock @harshavamsi good for me 🎉 thank you for allowing me contribute on this!
I'll open an issue and PR on https://github.com/yosefbs/aws-opensearch-connector to add a deprecated on the Readme and forward to the AwsSigV4 docs here once it's released.

USER_GUIDE.md
```diff
@@ -0,0 +1,171 @@
...
+  ssl: {
+    ca: fs.readFileSync(ca_certs_path),
+    // You can turn off certificate verification (rejectUnauthorized: false) if you're using self-signed certificates with a hostname mismatch.
+    // cert: fs.readFileSync(client_cert_path),
+    // key: fs.readFileSync(client_key_path)
+  },
+});
+```
+
+### To authenticate with [Amazon OpenSearch Service](https://aws.amazon.com/opensearch-service/) using AwsSigv4Signer
```
dblock, line 40: Nit: I would make it shorter, e.g. "Authenticate with Amazon OpenSearch" and remove the link from here or mentions of AwsSigv4Signer in the title.

USER_GUIDE.md
```diff
@@ -0,0 +1,171 @@
...
+    // With AWS SDK V2, Credentials.refreshPromise is used when available to refresh the credentials.
+
+    // Example with AWS SDK V3:
+    getCredentials: () => {
+      // Any other method to acquire a new Credentials object can be used.
+      const credentialsProvider = defaultProvider();
+      return credentialsProvider();
+    },
+
+    // Example with AWS SDK V2:
```
dblock, line 64: Separate the examples in separate code blocks. People copy paste a lot.

USER_GUIDE.md
```diff
@@ -0,0 +1,171 @@
...
+const client = new Client({
+  ...AwsSigv4Signer({
+    region: 'us-east-1',
+    // Must return a Promise that resolve to an AWS.Credentials object.
+    // This function is used to acquire the credentials when the client start and
+    // when the credentials are expired.
+    // The Client will refresh the Credentials only when they are expired.
+    // With AWS SDK V2, Credentials.refreshPromise is used when available to refresh the credentials.
+
+    // Example with AWS SDK V3:
```
dblock, line 57: What's the default? Do I need to always implement `getCredentials()`?

USER_GUIDE.md
```diff
@@ -0,0 +1,171 @@
...
+      },
+    },
+  };
+
+  var response = await client.indices.create({
+    index: index_name,
+    body: settings,
+  });
+
+  console.log('Creating index:');
```
dblock, line 100: Shouldn't console.log for what's going to happen be above the code (`var response = ...`) going to happen?

lib/aws/AwsSigv4Signer.js
```diff
@@ -0,0 +1,124 @@
...
+      // AWS SDK V2, needsRefresh should be available.
+      else if (typeof currentCredentials.needsRefresh === 'function') {
+        expired = currentCredentials.needsRefresh();
+      }
+      // AWS SDK V2, alternative to needsRefresh.
+      else if (currentCredentials.expired === true) {
+        expired = true;
+      }
+      // AWS SDK V2, alternative to needsRefresh and expired.
+      else if (currentCredentials.expireTime && currentCredentials.expireTime < new Date()) {
```
dblock, line 67: < or <=?

test/unit/awssigv4signer.test.js
```diff
@@ -0,0 +1,407 @@
+/*
```
dblock, line 1: This should match the file structure and naming in `lib`, so `test/unit/lib/aws/AwsSigv4Signer.test.js`

test/unit/awssigv4signer.test.js
```diff
@@ -0,0 +1,407 @@
...
+ */
+const { test } = require('tap');
+const { URL } = require('url');
+const { v4: uuidv4 } = require('uuid');
+const AwsSigv4Signer = require('../../lib/aws/AwsSigv4Signer');
+const AwsSigv4SignerError = require('../../lib/aws/errors');
+const { Connection } = require('../../index');
+const { Client, buildServer } = require('../utils');
+
+test('Sign with SigV4', (t) => {
```
dblock, line 19: Add a test for the absolute minimal syntax, aka with default signer options/credentials.

dblock (changes requested):
I have some more asks around tests, easy stuff.

USER_GUIDE.md
rawpixel-vincent: yes there is no default, getCredentials is required.
if not given an error is thrown, it's one of the test case.
I think the wording above make that clear, but it seems not, how can we update it to make it clear that `getCredentials()` is required?

USER_GUIDE.md
rawpixel-vincent: fair enough

USER_GUIDE.md
rawpixel-vincent: ok

USER_GUIDE.md
dblock: Maybe we can provide a default version?

lib/aws/AwsSigv4Signer.js
rawpixel-vincent: I think it's better to be strict in that case for the following reasons:
-previous library would renew the credentials on every requests as now we are refreshing the credentials only when needed
- this condition will probably never been triggered, issue is with the AWS.Credentials object definition that gives 3 different way to test if a Credentials object is expired
- mainly because under heavy load it's possible that between the time we check the request and the time the request is made the Credentials would be expired, so although that condition seems strict, it isn't.

I think it's good like this, but if it's blocker, maybe we can think of exposing an option that gives that choice to the user

something like:
```
new Client(
  ...AwsSigV4({
    ...,
    timeMarginBeforeRenewalMs: 1000 * 60
  })
)
// then that condition would be something like
else if (currentCredentials.expireTime && currentCredentials.expireTime < new Date(Date.now() - (opts.timeMarginBeforeRenewalMs || 0)) {
```
fwiw, I'm good with the current implementation until proven it's an issue, some feedback from users would help to find out.

test/unit/awssigv4signer.test.js
rawpixel-vincent: makes sense

test/unit/awssigv4signer.test.js
rawpixel-vincent: I think that would be `test('Basic aws (promises)', (t) => {` line 96 and `test('Basic aws (callback)', (t) => {` line 288
those are using the basic options/credentials

rawpixel-vincent:
@harshavamsi I wait for your feedback on the review from @dblock and then let me know if you want to address it or leave it to me 🙏🏻

USER_GUIDE.md
rawpixel-vincent: I'm not sure what you mean, those two examples are not meant to be copy/pasted, although they would work in the majority of case if people do copy/paste them.
I've tried to make it clear in the comment above how `getCredentials()` should be implemented.
I think it is good enough, although as discussed we can make it more clear by splitting the code example for v2 and v3.

USER_GUIDE.md
rawpixel-vincent: this is not in the scope of this PR.
Code examples can probably be improved, but might be better to open a specific issue for that.

USER_GUIDE.md
rawpixel-vincent: okay I got you, in my opinion giving a default is not the best idea as this is the kind of thing that should be thought out, and it could lead to confusion for consumer using different version of the sdk or just expecting this library to handle the credentials the way they think it should.
a default getCredentials() method could look very different between implementations and there would be good arguments for every implementation as why it should be the default.

rawpixel-vincent: that would also mean adding aws-sdk as a peer-dependency / optional-dependency or something like that.. I think it's best to leave it to the users that implement it.

lib/aws/AwsSigv4Signer.js
rawpixel-vincent: I don't think we should bother with this unless we have some feedback from aws sdk V2 users.
From the aws docs this condition will not be triggered. But because the Credentials object gives those 3 props, for the sake of following the specs it makes sense to test all possible properties.
```
      // AWS SDK V2, needsRefresh should be available.
      else if (typeof currentCredentials.needsRefresh === 'function') {
        expired = currentCredentials.needsRefresh();
      }
      // AWS SDK V2, alternative to needsRefresh.
      else if (currentCredentials.expired === true) {
        expired = true;
      }
      // AWS SDK V2, alternative to needsRefresh and expired.
      else if (currentCredentials.expireTime && currentCredentials.expireTime < new Date()) {
        expired = true;
      }
```
If the Credentials object is consistent in aws sdk v2, `Credentials.needsRefresh()` will always be used.

that point you made and my answer is still relevant for aws sdk v3, in v3 the Credentials object expose only `Credentials.expiration` to test if the credentials are expired.
With a Credentials object from aws sdk V3, this is the only condition that will execute
```
       // AWS SDK V3, Credentials.expiration is a Date object
      else if (currentCredentials.expiration && currentCredentials.expiration < new Date()) {
        expired = true;
      }
```

fwiw, this is the full block that check if the credentials have expired (or haven't been acquired yet)

```
      const currentCredentials = credentialsState.credentials;
      let expired = false;
      if (!currentCredentials) {
        // Credentials haven't been acquired yet.
        expired = true;
      }
      // AWS SDK V2, needsRefresh should be available.
      else if (typeof currentCredentials.needsRefresh === 'function') {
        expired = currentCredentials.needsRefresh();
      }
      // AWS SDK V2, alternative to needsRefresh.
      else if (currentCredentials.expired === true) {
        expired = true;
      }
      // AWS SDK V2, alternative to needsRefresh and expired.
      else if (currentCredentials.expireTime && currentCredentials.expireTime < new Date()) {
        expired = true;
      }
      // AWS SDK V3, Credentials.expiration is a Date object
      else if (currentCredentials.expiration && currentCredentials.expiration < new Date()) {
        expired = true;
      }
```

harshavamsi:
@rawpixel-vincent @dblock I've made changes where you've requested them. I've addressed the comments as well.

test/unit/awssigv4signer.test.js
harshavamsi: Yes, this would be the basic case. I agree.

lib/aws/AwsSigv4Signer.js
harshavamsi: This explanation looks good to me. Looks like the built-in functions from the library can handle this.

USER_GUIDE.md
harshavamsi: I agree with @rawpixel-vincent . This allows the user to use the library and implementation of his choice. Looking at community driven implementations of sigv4 there are users using both types of implementations and aws sdk versions.

lib/aws/AwsSigv4Signer.js
dblock: 👍

dblock (approved):

dblock:
I merged this, great work!

dblock:
I opened https://github.com/opensearch-project/opensearch-js/issues/290 for a possible improvement.

rawpixel-vincent:
@harshavamsi @dblock the tests are not running after they been moved to lib/aws see #294

Issue #291, opened 2022-09-07
[BUG] Connection headers mutability and unintended reuse of headers


mmtdm:
### What is the bug?
[Connection.js headers property](https://github.com/opensearch-project/opensearch-js/blob/main/lib/Connection.js#L221) is reused over multiple requests, and due to mutability of headers, this can cause unwanted headers to be sent with subsequent requests. 

### How can one reproduce the bug?
Using opensearch-project/opensearch with aws-opensearch-connector, call search repeatedly from multiple expressjs processes, being sure to execute queries on es cluster that run for more than 5 minutes (Ultrawarm index in our case).

Eventually, an X-Amz-Date header will be added to Connection.headers, and due to [logic in aws4.js](https://github.com/mhart/aws4/blob/master/aws4.js#L157), this header will be sent with every request, and never get updated.

Eventually, the permanent X-Amz-Date header will expire, and all subsequent requests will fail with "Signature expired" error.

The issue only resolves itself when you recreate a new Client (which recreates a new Connection). Eventually, the header will be rewritten though, and the process will repeat itself.

### What is the expected behavior?
headers should not be reused across multiple distinct requests. Connection class should not use a reference to this.headers when initializing a request, it should make a copy: 

`headers: Object.assign({}, this.headers), // create a clone`

### What is your host/environment?
Linux/node 16

### Do you have any additional context?
This issue seems to only arise in the presence of multiple requests being handled for an es cluster that has queued search threads and long running query executions that exceed 5 minutes.

Is there an initialization or usage pattern that would result in ability to pass in blank headers to each request? Otherwise, it really does seem like the Connection headers property is being passed around by reference but also reused over multiple requests.

harshavamsi:
Hi, @mmtdm opensearch-js now supports [sigv4 authentication]( https://github.com/opensearch-project/opensearch-js/blob/main/USER_GUIDE.md#using-aws-v3-sdk) support out of the box. We've enhanced the implementation over how requests are made over the transport layer and this takes into consideration credentials expiring. Can you give this a try and let us know if this helps your case?

mmtdm:
Thank you kindly @harshavamsi I will try your recommendation and report on my findings. 

[...]

PR #292, opened 2022-09-07
issue #291 initialize request.headers by value, not reference

mmtdm:
### Description
This modifies how request.headers are initialized. Instead of initializing request.headers by reference to Connection.headers, this makes a copy of Connection.headers. This way, any changes to request.headers are not reflected in the underlying connection, which is reused for multiple requests.

### Issues Resolved
https://github.com/opensearch-project/opensearch-js/issues/291

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -218,7 +218,7 @@ class Connection {
       origin: url.origin,
       // https://github.com/elastic/elasticsearch-js/issues/843
       port: url.port !== '' ? url.port : undefined,
-      headers: this.headers,
+      headers: Object.assign({}, this.headers), // issue #291 initialize by value, not reference
       agent: this.agent,
     };
 


mmtdm:
cc: @dblock

harshavamsi:
@mmtdm can you amend your commit with `-s` so the DCO check can pass?

harshavamsi:
LGTM! @dblock @VachaShah

dblock (changes requested):
I think we need a unit test for this, or we'll accidentally break this behavior again in the future. Please add?

lib/Connection.js
```diff
@@ -218,7 +218,7 @@ class Connection {
       origin: url.origin,
       // https://github.com/elastic/elasticsearch-js/issues/843
```
dblock, line 219: I dislike comments like these. I say let's add a unit test and we don't need to explain.

mmtdm: agreed. Also, original fork for this PR was deleted, so I'm going to close this PR and open a new one, and will add a unit test.

PR #294, opened 2022-09-09
fix awssigv4signer.test.js tests not running

rawpixel-vincent:
### Description
Fix awssigv4signer.test.js tests not running after they have been moved to test/unit/lib/aws


dblock (approved):

package.json
```diff
@@ -34,16 +34,16 @@
...
+    "test:unit": "tap test/unit/{*,**/*,**/**/*}.test.js",
     "test:acceptance": "tap test/acceptance/*.test.js",
     "test:integration": "node test/integration/index.js",
     "test:integration:helpers": "tap test/integration/helpers/*.test.js",
     "test:integration:helpers-secure": "tap test/integration/helpers-secure/*.test.js",
     "test:types": "tsd",
-    "test:coverage-90": "tap test/{unit,acceptance}/{*,**/*}.test.js --coverage --branches=90 --functions=90 --lines=90 --statements=90 --nyc-arg=\"--exclude=api\"",
-    "test:coverage-report": "tap test/{unit,acceptance}/{*,**/*}.test.js --coverage --branches=90 --functions=90 --lines=90 --statements=90 --nyc-arg=\"--exclude=api\" && nyc report --reporter=text-lcov > coverage.lcov",
-    "test:coverage-ui": "tap test/{unit,acceptance}/{*,**/*}.test.js --coverage --coverage-report=html --nyc-arg=\"--exclude=api\"",
+    "test:coverage-90": "tap test/{unit,acceptance}/{*,**/*,**/**/*}.test.js --coverage --branches=90 --functions=90 --lines=90 --statements=90 --nyc-arg=\"--exclude=api\"",
```
dblock, line 44: Is there a way to simplify all this stuff and just do `test/....everything*`? What else is in that directory? 

Half the time I accidentally call a file `foobar.js` instead of `foobar.test.js` and it's simply ignored and I think the test is passing.

rawpixel-vincent: I see, I'm not completely aware of the structure of this, so i think this could be addressed by someone more akin of the tests workflow. 
fwiw this doesn't seems to break anything for now

VachaShah (approved):

Issue #268, opened 2022-07-27
[BUG] Integration tests are failing intermittently


dblock:
### What is the bug?

See https://github.com/opensearch-project/opensearch-js/runs/7454554035?check_suite_focus=true, currently failing on main.

### What is the expected behavior?

Tests to pass.

### Do you have any additional context?

Red herring in https://github.com/opensearch-project/opensearch-js/pull/266#issuecomment-1195960934

kavilla:
Re-ran it and success. Indicating that these tests are flaky, but they've seem pretty consistent. So might have been a fluke are some availability issue.

nhtruong:
I've successfully recreated this error on my local env!
Since all flaky workflows run integration tests, my focus were on the 4 integration helper tests. All of these tests read from `stackoverflow.ndjson` as a stream, which is similar to [this issue on node-tap repo](https://github.com/tapjs/node-tap/issues/422). Intentionally making the stream fail (like giving it a wrong path) will result in identical error message that we've seen on the failed jobs:
```bash
theotr@88665a51fc3f opensearch-js % tap test/integration/helpers-secure/search.test.js -C
TAP version 13
(node:9350) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
(Use `node --trace-warnings ...` to show where the warning was created)
# Subtest: test/integration/helpers-secure/search.test.js
    1..0 # no tests found
not ok 1 - test/integration/helpers-secure/search.test.js # time=30010.195ms
  ---
  env: {}
  file: test/integration/helpers-secure/search.test.js
  timeout: 30000
  command: /usr/local/Cellar/node/18.9.1/bin/node
  args:
    - test/integration/helpers-secure/search.test.js
  stdio:
    - 0
    - pipe
    - 2
  cwd: /Users/theotr/IdeaProjects/opensearch-js
  exitCode: null
  signal: SIGTERM
  ...

1..1
# failed 1 test
# time=30043.437ms
```

nhtruong:
It is actually the OS Instance in the container that caused this issue, even though an error in the file stream will throw the exact same timeout/no-tests error. Configurations that failed had the following printout from the container's logs:
```bash
docker logs opensearch_opensearch_1
Killing opensearch process 10
Killing performance analyzer process 11
```
```bash
curl localhost:9200/_cat/health
[...]

PR #304, opened 2022-10-03
Retry for OpenSearch Container

nhtruong:
### Description
Tests that run `docker-compose up` randomly fail with "no tests found error". Rerunning the failed jobs resolves the issue. This is [a bug in Opensearch build](https://github.com/opensearch-project/opensearch-build/issues/2568) that causes the container to intermittently fail. Restarting the container when encountering this issue is a workaround this PR is addressing.

### Issues Resolved

closes https://github.com/opensearch-project/opensearch-js/issues/268

### DIFF
--- a/.ci/opensearch/Dockerfile.opensearch
+++ b/.ci/opensearch/Dockerfile.opensearch
@@ -5,4 +5,10 @@ ARG opensearch_path=/usr/share/opensearch
 ARG opensearch_yml=$opensearch_path/config/opensearch.yml
 
 ARG SECURE_INTEGRATION
+
+HEALTHCHECK --start-period=20s --interval=5s --retries=2 --timeout=1s \
+  CMD if [ "$SECURE_INTEGRATION" != "true" ]; \
+  then curl --fail localhost:9200/_cat/health; \
+  else curl --fail -k https:/localhost:9200/_cat/health -u admin:admin; fi
+
 RUN if [ "$SECURE_INTEGRATION" != "true" ] ; then $opensearch_path/bin/opensearch-plugin remove opensearch-security; fi
--- a/.ci/opensearch/docker-compose.yml
+++ b/.ci/opensearch/docker-compose.yml
@@ -2,6 +2,7 @@ version: '3'
 
 services:
   opensearch:
+    restart: always
     build:
       context: .
       dockerfile: Dockerfile.opensearch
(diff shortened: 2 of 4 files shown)


dblock:
Maybe we can either fix the underlying issue or use something like https://www.npmjs.com/package/retry instead of rolling our own?

nhtruong:
@dblock I'm looking into finding/fixing the source myself. I'm just checking if we can use retry to ease the pain of manually retrying failed jobs. Unfortunately, the retry I implemented in the first commit will still fail all 3 attempts if the first attempt fails. Now I'm using this PR to test out different strategies. My current suspicion is that TAP is somehow looking for TS or JSX files as if `TAP_TS` and `TAP_TXS` were set to 1.

nhtruong:
Found the cause and a workaround. Please check the updated description for details.

dblock:
I hate it but I think it works ;)

dblock (approved):

dblock:
Berfore we merge this, is there a way to retry the tests at GHA level? I'm worried about hacks on top of hacks.

harshavamsi:
Could look into https://github.com/marketplace/actions/retry-action and https://github.com/marketplace/actions/retry-step. Validate container start up docker and execute retry logic.

nhtruong:
> Could look into https://github.com/marketplace/actions/retry-action and https://github.com/marketplace/actions/retry-step. Validate container start up docker and execute retry logic.

I thought of retry-step for the Container Startup step but decided not to use it:
- It will retry the whole step while we only need to restart the container instead of going through the entire clean/build/boot-up sequence.
- This step never failed (or failed silently), so we will still have to modify this step to grep for the kill signals to call `exit 1` explicitly to signal retry-step to retry.
- The fix is the make file, people running make commands on their local machine will benefit from this too, not just github workflow.

nhtruong:
@dblock @harshavamsi 
Sooo the best the solution is stupidly simple! The `docker-compose up` step not failing on the failed tests led me to believe that the container was still running and only the OS service inside the container that crashed. However, what actually happened is, the container itself crashed soon after `docker-compose up`. And we just need to tell docker to always restart the container (instead of the default `no`) when it crashes.

https://hub.docker.com/r/willfarrell/autoheal/ is not appropriate for this situation because it's mean for crashed services, not crashed containers.

nhtruong:
Testing a few more times to make sure that the problem is solved.

nhtruong:
I spoke too soon. This bug doesn't always crash the container when the opensearch processed is killed. It only does that sometimes.

dblock:
@nhtruong This is pretty clean retry-wise. Do you want to merge it or it doesn't work?

nhtruong:
@dblock I've tested this on over 700 runs and non failed. So, it's good to go!

VachaShah (approved):

PR #315, opened 2022-10-25
Exclude dependabot from changelog-enforcer

nhtruong:
Signed-off-by: Theo Truong <theotr@amazon.com>


dblock (approved):

nhtruong:
Going to implement something like [this](https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/dependabot_pr.yml) instead.

PR #366, opened 2023-01-20
Added Support for AOSS

### LINKED ISSUE #356: [FEATURE] Add support for OpenSearch Serverless

With https://github.com/opensearch-project/opensearch-js/pull/333 one can supply `aoss`. I believe `client.info()` will be broken because that doesn't exist in serverless. What else?

Document support for OpenSearch Serverless in the compatibility matrix, add integration tests.

Coming from https://github.com/opensearch-project/opensearch-clients/issues/44

nhtruong:
Signed-off-by: Theo Truong <theotr@amazon.com>

### Issues Resolved
Closes #356 


lib/aws/AwsSigv4Signer.js
```diff
@@ -77,6 +77,12 @@ function AwsSigv4Signer(opts = {}) {
     request.region = opts.region;
     request.headers = request.headers || {};
     request.headers['host'] = request.hostname;
+    if (opts.service === 'aoss') {
```
dblock, line 80: I am strongly opposed to this. The server is aligning the implementation.

dblock (changes requested):

lib/aws/AwsSigv4Signer.js
nhtruong: Adding x-ams-context-sha256 works for both :)

wbeckler: Amazing!!

dblock:
Nicely done @nhtruong. Add a test for the added header and we're good to go?

nhtruong:
@dblock Yes. I've tested and it's working for both. the `x-ams-context-sha256` header is optional for managed service but required by AOSS. Adding it makes it work for both services for requests with and without a body.

test/unit/lib/aws/awssigv4signer.test.js
```diff
@@ -50,6 +50,7 @@ test('Sign with SigV4', (t) => {
   const signedRequest = auth.buildSignedRequestObject(request);
   t.hasProp(signedRequest.headers, 'X-Amz-Date');
   t.hasProp(signedRequest.headers, 'Authorization');
+  t.hasProp(signedRequest.headers, 'x-amz-content-sha256');
```
dblock, line 53: Is the value always the same? Maybe add a test that ensures this value is what you expect it to be?

dblock (approved):

lib/aws/AwsSigv4Signer.js
```diff
@@ -77,7 +78,12 @@ function AwsSigv4Signer(opts = {}) {
     request.region = opts.region;
     request.headers = request.headers || {};
     request.headers['host'] = request.hostname;
-    return aws4.sign(request, credentialsState.credentials);
+    const signed = aws4.sign(request, credentialsState.credentials);
+    signed.headers['x-amz-content-sha256'] = crypto
+      .createHash('sha256')
+      .update(request.body || '', 'utf8')
+      .digest('hex');
```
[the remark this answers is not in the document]
mhart: Ah, my bad, only does it for S3. https://github.com/mhart/aws4/blob/master/aws4.js#L158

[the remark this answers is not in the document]
nhtruong: That's what we thought too but it actually does not add `x-amz-content-sha256` header, which is required by AOSS It's optionally for Managed Service, which is why we didn't need it before.

dblock:
@nhtruong Care to do https://github.com/opensearch-project/opensearch-js/issues/371 as well? thx

PR #367, opened 2023-01-20
Fix serverless signing with body

### LINKED ISSUE #356: [FEATURE] Add support for OpenSearch Serverless

harshavamsi:
Signed-off-by: Harsha Vamsi Kalluri <harshavamsi096@gmail.com>

### Description
Proof of concept with SigV4 signing for serverless. This is still a work in progress. I can't seem to understand why this does not work. I compute the body hash and set the `x-amz-content-sha256` header and use that to sign the request, but there is a mismatch in the signature.

### Issues Resolved
Closes #356 


lib/aws/AwsSigv4Signer.js
```diff
@@ -77,13 +78,23 @@ function AwsSigv4Signer(opts = {}) {
     request.region = opts.region;
     request.headers = request.headers || {};
     request.headers['host'] = request.hostname;
+    request.headers['X-Amz-Content-Sha256'] = hash(request.body || '', 'hex');
```
dblock, line 81: Looks like this was a bit more involved, https://github.com/opensearch-project/opensearch-js/pull/366

lib/aws/AwsSigv4Signer.js
```diff
@@ -77,13 +78,23 @@ function AwsSigv4Signer(opts = {}) {
...
+    request.extraHeadersToIgnore = {
+      'content-length': true,
+    };
```
dblock, line 84: Was this the issue? Content-length needed to be included?

https://github.com/opensearch-project/opensearch-js/pull/366/files

harshavamsi: No, looks like that PR is not including `x-amz-content-sha256` in the signing process. It signs the request first and then computes the hash in the header. Weird, I tried this but it did not work.

Issue #269, opened 2022-07-27
[FEATURE] Rewrite this client in TypeScript


dblock:
### Is your feature request related to a problem?

Coming from https://github.com/opensearch-project/opensearch-js/pull/266#issuecomment-1197061005, TypeScript bindings are maintained by hand. Development in pure JS is slower, potentially more error prone.

### What solution would you like?
Rewrite the client in TypeScript, get rid of pure JS code.

nhtruong:
To assure that the rewrite does not affect existing users of the JS client, while capturing all features of this JS client in the TS client, I propose the following steps:

1) Convert the current JS client into a TS client with minimal changes to the client code itself:
- Create a dev branch called `typescript-conversion` from `main`. All PRs into main from there on must also be reflected in the `typescript-conversion` branch while conversion, this first step, is happening.
- Rename all .js files to .ts files
- Add a Typescript workflow to compile TS into JS
- Update Node JS release logic
- Update Github Workflows if necessary
- Finally merge `typescript-conversion` into main.

2) Convert JS syntax into TS syntax for non-API functions/files

3) Write a generator for API functions (We can manually convert a few API functions to TS first to have a better idea of what the generated functions would look like)

4) Refactor the client to take full advantage of TS features (Note that the changes in step 2 are limited to the scope of each TS file that was previously renamed from JS)

nhtruong:
@timursaurus has already started the rewrote effort on his own [repo](https://github.com/timursaurus/opensearch-ts). It gives us an idea of what the final product would look like. @timursaurus can also port a lot of his work from that repo into this, while other contributors can also make PRs to contribute to this effort.

dblock:
@nhtruong This LGTM. I'd call the branch `feature/typescript-conversion` per convention with other repos.

RealZacharyFowler:
Hello is someone currently working on this? It looks like there but I saw there is no one assigned to it.

nhtruong:
@RealZacharyFowler you're absolutely welcomed to chip in. No one has made any PR just yet. Though do note that this is a rewrite of the entire client. We're not just doing a 1-to-1 translation to TS. We're looking to make big improvements to get rid of glaring issues in the current JS clients WHILE maintaining backward compatibility (meaning users should not have to make next to no major changes on their applications when switching over to our new TS client). This means we will have to thoroughly study the current JS client to be able to rewrite it in TS.

PR #378, opened 2023-02-03
 Allow fields in BulkOperation to be optional

camerondurham:
### Description

This should change the `BulkOperation` type to more closely match the OpenSearch API. It makes the API types for requests slightly less restrictive and easier to use in Typescript.

From my limited understanding of the parser is that not all fields are required for bulk requests. From the BulkRequestParserTests in the project there are a few examples showing this behavior, where only `_id` provided: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/action/bulk/BulkRequestParserTests.java#L119

From the (1.x and 2.x) API docs too, it doesn't look like all fields in `BulkOperation` type definition are required. 

https://opensearch.org/docs/1.2/opensearch/rest-api/document-apis/bulk/#request-body
https://opensearch.org/docs/2.5/api-reference/document-apis/bulk/#request-body
> All actions support the same metadata: `_index`, `_id`, and `_require_alias`. If you don’t provide an ID, OpenSearch generates one automatically, which can make it challenging to update the document at a later time.

If this was intentional to have all fields required, I'm curious to know why?

This might be more of an "issue" but the change appeared too straightforward for me to do that. 

Thanks!

### Issues Resolved
N/A

Signed-off-by: Cameron Durham <camdurha@amazon.com>

### DIFF
--- a/api/types.d.ts
+++ b/api/types.d.ts
@@ -40,12 +40,12 @@ export interface BulkIndexOperation extends BulkOperation { }
 export interface BulkIndexResponseItem extends BulkResponseItemBase { }
 
 export interface BulkOperation {
-  _id: Id;
-  _index: IndexName;
-  retry_on_conflict: integer;
-  routing: Routing;
-  version: VersionNumber;
-  version_type: VersionType;
+  _id?: Id;
+  _index?: IndexName;
+  retry_on_conflict?: integer;
+  routing?: Routing;
+  version?: VersionNumber;
+  version_type?: VersionType;
 }
 
 export interface BulkOperationContainer {


nhtruong:
@camerondurham Thanks for catching that! Would you mind updating the ChangeLog to log your change as well?

dblock:
I really think we should have tests for this, if anything just to exercise these types?

camerondurham:
That's fair, definitely makes sense to test Typescript declarations in some way.

I can work on a test that creates some instance types from the TS declarations. Let me know if you instead meant actually testing the TS bindings (e.g. writing some of the JS tests in TS to test these bindings) which might take more time/duplicate work but might add some more confidence too.

dblock:
Up to you @camerondurham, I'm also ok to merge as is.

Note that we also have #269 that might be the real solution.

dblock (approved):

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Dependencies
 ### Changed
 - Remove test artifacts from gh_pages workflow ([#335](https://github.com/opensearch-project/opensearch-js/issues/335))
+- Make fields in `BulkOperation` optional to match OpenSearch Bulk API requirements ([#378](https://github.com/opensearch-project/opensearch-js/pull/378))
```
dblock, line 40: I would re-add an empty line here ;)

camerondurham:
Yeah I totally agree, a TS rewrite would be ideal. Maybe those type declaration tests could be one of many subtasks for the TS client?

But if you're okay to merge, we can close this PR out and I can send a separate one for TS type declaration tests. :)

Also, appreciate the fast responses on this PR from the team!

dblock (approved):

dblock:
Will merge. We're just waiting for someone brave (wink wink) to undertake a rewrite ;)

nhtruong (approved):

PR #423, opened 2023-03-09
[CCI] Remove waitCluster in Integration Tests

### LINKED ISSUE #422: [PROPOSAL] Remove waitCluster in Integration Tests
Every integration test starts of by hitting [waitCluster](https://github.com/opensearch-project/opensearch-js/blob/66c3a99cb04efe1ae6607a4fc76ef04317fd3aec/test/utils/index.js#L40-L53) method, which waits for the cluster to be up and ready before the test is executed. This is redundant because we have already implemented [healthcheck](https://github.com/opensearch-project/opensearch-js/blob/main/.ci/opensearch/Dockerfile.opensearch#L9-L12) along with [autoheal](https://github.com/opensearch-project/opensearch-js/blob/main/.ci/opensearch/docker-compose.yml#L19-L28) to assure that the OS cluster is up and running before the integration workflow executes these tests.

Moreover, `waitCluster` is waiting for status of `green`, which is usually not applicable for local testing because:
Since we're testing the client functionality (mostly API endpoints), the OS cluster only has a single node (for simplicity and it's sufficient for development and testing). However, by default every index created requires 1 replica, which means it needs a 2nd node to deploy a backup copy to. This causes the index to have status of `yellow`, which leads to `waitCluster` to wait forever if there's any index already exists on the OS cluster. For most developers, it's normal to have a few existing indices on the local OS cluster for development. But we have to delete these indices or change their replicas_count to 0 if we want to run the integration tests locally.

_Note that having a yellow status on single-node cluster is normal. The cluster is still working as intended. You just dont have a backup copy of the indices. I would say this is a bug from the server standpoint: Indices shouldn't have number of replicate set to 1 by default if there's only 1 node in the cluster._

In other words, we should remove `waitCluster` because:
- It's redundant
- It makes it harder to run integration tests locally

Wielmany:
### Description
Removing waitCluster in Integration Tests due to it's redundancy and complication of running integration tests locally
### Issues Resolved
closes #422  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Add release details to releasing.md ([319](https://github.com/opensearch-project/opensearch-js/pull/319))
 - Added Amazon OpenSearch Serverless in user_guide ([#372](https://github.com/opensearch-project/opensearch-js/issues/372))
+- Remove waitCluster in Integration Tests ([#422](https://github.com/opensearch-project/opensearch-js/issues/422))
 ### Dependencies
 - Bumps `aws4` from 1.11.0 to 1.12.0
 - Bumps `minimist` from 1.2.6 to 1.2.8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Add release details to releasing.md ([319](https://github.com/opensearch-project/opensearch-js/pull/319))
 - Added Amazon OpenSearch Serverless in user_guide ([#372](https://github.com/opensearch-project/opensearch-js/issues/372))
-- Remove waitCluster in Integration Tests ([#422](https://github.com/opensearch-project/opensearch-js/issues/422))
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Add release details to releasing.md ([319](https://github.com/opensearch-project/opensearch-js/pull/319))
 - Added Amazon OpenSearch Serverless in user_guide ([#372](https://github.com/opensearch-project/opensearch-js/issues/372))
+- Remove waitCluster in Integration Tests ([#422](https://github.com/opensearch-project/opensearch-js/issues/422))
```
nhtruong, line 8: This should be in the `Removed` section.

test/integration/helpers-secure/search.test.js
```diff
@@ -30,7 +30,7 @@ const client = new Client({
 });
 
 beforeEach(async () => {
-  await waitCluster(client);
+
```
nhtruong, line 33: Linter failed becaused of this blank line.

nhtruong:
Thanks @Wielmany!
Remember to add `closes #422` (instead of just #422) in the description of the PR so that when the PR is merged, the ticket is closed as well :)

Issue #365, opened 2023-01-20
Deprecation warning when using Node 16 and Typescript


dibell:
When using Node 16 I see a deprecation warning when using the ResponseError exported from a subpath.

`[DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at ./opensearchplay/node_modules/@opensearch-project/opensearch/package.json.
Update this package.json to use a subpath pattern like "./*".`

Here is a short example to reproduce this when run with ts-node:

```typescript
import { Client } from '@opensearch-project/opensearch';
import { ResponseError } from '@opensearch-project/opensearch/lib/errors';

const client = new Client({ node: 'http://admin:admin@localhost:9200' });

const describeIndex = async () => {
    try {
        const response = await client.indices.get({ index: 'doesnotexist' });
        console.log(response.body);
    } catch (err) {
        if (err instanceof ResponseError) console.log(err.message, err.statusCode, err.meta.body.result);
        else console.log(err);
    }
};

describeIndex().catch((err) => console.error(err));
```

I guess either update the package.json exports to .* as suggested in the warning or export ResponseError from the root?

TIA

saga-dash:
I encountered the same issue, and in my environment, the following solution worked:

```
import { errors } from '@opensearch-project/opensearch';

if (error instanceof errors.ResponseError) {
  // Handle the error
}
```

Issue #247, opened 2022-05-24
[Clean] remove npm in workflows


ananzh:
Since we are using yarn, we should make consistency by switching npm to yarn. No need for two package managers in one single project.

PR #424, opened 2023-03-10
[#365] Deprecation warning when using Node 16 fixed by changing expor…

frost017:
…t in package.json Signed-off-by: Raiymbek Aiymbet <r.aiymbet@gmail.com>

### Description
This minor change removes deprecation warning when using Node 16 and TypeScript. The fix is about changing **export** block in package.json
_```(node:31064) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at /node_modules/@opensearch-project/opensearch/package.json imported from /index.js.
Update this package.json to use a subpath pattern like "./*".
(Use `node --trace-deprecation ...` to show where the warning was created)```_

### Issues Resolved
This fixes following problem: Deprecation warning when using Node 16 and TypeScript https://github.com/opensearch-project/opensearch-js/issues/365.


nhtruong:
We're only using `yarn`, not `npm`. So the addition of `package-lock.json` will lead to confusion.

frost017:
> We're only using `yarn`, not `npm`. So the addition of `package-lock.json` will lead to confusion.

So this issue and PR is not relevant? Should I remove npm then?(There is issue about removing npm https://github.com/opensearch-project/opensearch-js/issues/247)

nhtruong:
> > We're only using `yarn`, not `npm`. So the addition of `package-lock.json` will lead to confusion.
> 
> So this issue and PR is not relevant? Should I remove npm then?(There is issue about removing npm #247)

did you run `yarn install` or `npm install` or both? Running `npm install` will generate `package-lock` file

Issue #240, opened 2022-05-20
[eslint] duplicate conditions in indices and nodes API


ananzh:
**What is the bug?**
eslint shows two APIs have duplicate conditions.

```
yarn lint
yarn run v1.22.5
$ eslint .

/opensearch-js/api/api/indices.js
  1112:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  1115:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  1118:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  1127:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if

/opensearch-js/api/api/nodes.js
  169:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  180:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  183:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if

✖ 7 problems (7 errors, 0 warnings)
```

**How can one reproduce the bug?**
**After eslint PR is merged**
1. Go to '.eslintrc.json'
2. Remove ' "no-dupe-else-if": "off"'
3. Run `yarn lint`

**What is the expected behavior?**
Research on this
Remove duplicate conditions

ananzh:
duplication:
https://github.com/opensearch-project/opensearch-js/blob/5328c314608dda8d783799d32be3fe927ae1e935/api/api/nodes.js#L116-L131

frost017:
Good day @ananzh !
I want to let you know that I am working on this issue. Thank you for this opportunity!

nhtruong:
Closing since this has been replaced by generated API functions.

PR #428, opened 2023-03-11
[CCI][#240][eslint] Fixed duplicate conditions in indices and nodes API

frost017:
### Description
**Summary about** `indices.js`: This error is triggered when running `yarn lint` with next rule removed - `"no-dupe-else-if": "off"`. Previously, there were else-if statements having the same condition, which means only one of them can be executed and others are ignored. These were removed because of duplication:
```
else if (index != null && type != null) {
  if (method == null) method = 'PUT';
  path = '/' + encodeURIComponent(index) + '/' + encodeURIComponent(type) + '/' + '_mapping';
} else if (index != null && type != null) {
  if (method == null) method = 'PUT';
  path = '/' + encodeURIComponent(index) + '/' + '_mapping' + '/' + encodeURIComponent(type);
} else if (index != null && type != null) {
  if (method == null) method = 'PUT';
  path = '/' + encodeURIComponent(index) + '/' + encodeURIComponent(type) + '/' + '_mappings';
} else if (index != null && type != null) {
  if (method == null) method = 'PUT';
  path = '/' + encodeURIComponent(index) + '/' + '_mappings' + '/' + encodeURIComponent(type);
}
```
The code above resulted in following:
```
  1572:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  1575:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  1578:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  1587:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
```
**FIX** in `indices.js`: Removed duplicate conditions and made separate `if` statement for each case. By doing this, path is picked correctly and cases are not ignored because of duplicate conditions.

**Summary about** `nodes.js`: Same case as in `indices.js`. It has duplicate conditions, all check same case. It means all four are executed, which makes three conditions to be ignored. These lines of code were removed:
```
if ((node_id || nodeId) != null) {
  if (method == null) method = 'GET';
  path = '/' + '_nodes' + '/' + encodeURIComponent(node_id || nodeId) + '/' + 'hot_threads';
} else if ((node_id || nodeId) != null) {
  if (method == null) method = 'GET';
  path =
    '/' +
    '_cluster' +
    '/' +
    'nodes' +
    '/' +
    encodeURIComponent(node_id || nodeId) +
    '/' +
    'hotthreads';
} else if ((node_id || nodeId) != null) {
  if (method == null) method = 'GET';
  path = '/' + '_nodes' + '/' + encodeURIComponent(node_id || nodeId) + '/' + 'hotthreads';
} else if ((node_id || nodeId) != null) {
  if (method == null) method = 'GET';
  path =
    '/' +
    '_cluster' +
    '/' +
    'nodes' +
    '/' +
    encodeURIComponent(node_id || nodeId) +
    '/' +
    'hot_threads';
}
```
The code above resulted in following:
```
  189:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  200:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
  203:14  error  This branch can never execute. Its condition is a duplicate or covered by previous conditions in the if-else-if chain  no-dupe-else-if
```

**FIX** in `nodes.js`: Removed those aforementioned four conditions, and put one `if-else` condition which checks if `node_id` or `nodeId` is not `null`.

### Issues Resolved
**Following issue is resolved:** https://github.com/opensearch-project/opensearch-js/issues/240

### DIFF
--- a/api/api/indices.js
+++ b/api/api/indices.js
@@ -1554,57 +1554,6 @@ IndicesApi.prototype.putIndexTemplate = function indicesPutIndexTemplateApi(
  * @returns {{abort: function(), then: function(), catch: function()}|Promise<never>|*}
  */
 
-// Issue #240. This commented section must be removed, if the same function below this one is approved to be correct solution to the duplication error
-// IndicesApi.prototype.putMapping = function indicesPutMappingApi(params, options, callback) {
-//   [params, options, callback] = normalizeArguments(params, options, callback);
-
-//   // check required parameters
-//   if (params.body == null) {
-//     const err = new this[kConfigurationError]('Missing required parameter: body');
-//     return handleError(err, callback);
-//   }
-
-//   let { method, body, index, type, ...querystring } = params;
-//   querystring = snakeCaseKeys(acceptedQuerystring, snakeCase, querystring);
-
-//   let path = '';
-//   if (index != null && type != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + encodeURIComponent(index) + '/' + encodeURIComponent(type) + '/' + '_mapping';
-//   } else if (index != null && type != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + encodeURIComponent(index) + '/' + '_mapping' + '/' + encodeURIComponent(type);
-//   } else if (index != null && type != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + encodeURIComponent(index) + '/' + encodeURIComponent(type) + '/' + '_mappings';
-//   } else if (index != null && type != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + encodeURIComponent(index) + '/' + '_mappings' + '/' + encodeURIComponent(type);
-//   } else if (index != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + encodeURIComponent(index) + '/' + '_mapping';
-//   } else if (type != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + '_mappings' + '/' + encodeURIComponent(type);
-//   } else if (index != null) {
-//     if (method == null) method = 'PUT';
-//     path = '/' + encodeURIComponent(index) + '/' + '_mappings';
-//   } else {
-//     if (method == null) method = 'PUT';
-//     path = '/' + '_mapping' + '/' + encodeURIComponent(type);
-//   }
-
-//   // build request object
-//   const request = {
-//     method,
-//     path,
-//     body: body || '',
-//     querystring,
-//   };
-
-//   return this.transport.request(request, options, callback);
-// };
-
 IndicesApi.prototype.putMapping = function indicesPutMappingApi(params, options, callback) {
   [params, options, callback] = normalizeArguments(params, options, callback);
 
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -10,7 +10,9 @@
       "import": "./index.mjs"
     },
     "./aws": "./lib/aws/index.js",
-    "./": "./"
+    "./*": "./",
+    "./opensearch/*": "./lib/opensearch/*",
+    "./lib/*": "./lib/*"
```
dblock, line 15: Are these changes related to this fix? Rebase your branch.

package.json
```diff
@@ -91,6 +93,7 @@
     "xmlbuilder2": "^3.0.2"
   },
   "dependencies": {
+    "@types/react": "17.0.53",
```
dblock, line 96: I think you didn't want this change.

api/api/indices.js
```diff
@@ -1568,30 +1620,20 @@ IndicesApi.prototype.putMapping = function indicesPutMappingApi(params, options,
...
-    if (method == null) method = 'PUT';
-    path = '/' + encodeURIComponent(index) + '/' + '_mappings';
+    path = '/' + '_mappings' + '/' + encodeURIComponent(type) + '/';
   } else {
     if (method == null) method = 'PUT';
-    path = '/' + '_mapping' + '/' + encodeURIComponent(type);
+    path = '/' + '_mapping' + '/';
   }
 
+  path += '_mapping';
```
dblock, line 1635: This will make `/_maping/_mapping` in the last `else` case at least.

dblock (changes requested):
Thanks!

I am concerned that no tests broke given what I think is a bug below. Can you please beef up the `putMapping` tests to make sure we cover every path of the changed `if`? 

Delete commented out code, and any unrelated changes.

api/api/indices.js
frost017: `/_mapping/_mapping` shouldn't be the case in any way, right?

package.json
frost017: No, they aren't. It was intended to fix deprecation mapping error when using NodeJS 16 and TypeScript https://github.com/opensearch-project/opensearch-js/issues/365. I am removing this fix for now, leaving it as it was previously.

frost017: > Are these changes related to this fix? Rebase your branch.

No, they aren't. It was intended to fix deprecation mapping error when using NodeJS 16 and TypeScript https://github.com/opensearch-project/opensearch-js/issues/365. I am removing this fix for now, leaving it as it was previously.

frost017:
> I am concerned that no tests broke given what I think is a bug below. Can you please beef up the `putMapping` tests to make sure we cover every path of the changed `if`?

Yes, of course, I will do my best to cover it with tests. The thing is that I am not yet familiar with the structure of unit tests in the project, so I may put `getMapping` tests in the wrong place:)

Issue #206, opened 2022-03-03
[BUG] opensearchtypes is not consistent with our CamelCase naming convention


ananzh:
**What is the bug?**
When we fork elasticsearch-js, we followed the same naming rules as we fork opensearch and opensearch-dashboards. So `es` is changed to `opensearch`.

```
export * as estypes from './types'
export * as opensearchtypes from './types'
```

However `opensearchtypes` is not consistent with our CamelCase naming convention in osd.  

**What is the expected behavior?**
1) Change `opensearchtypes` [here](https://github.com/opensearch-project/opensearch-js/blob/68eea00b078710c876bd4c65f7b817c8d418a25f/api/new.d.ts#L698) to `opensearchTypes` and check the code base to make sure all `opensearchtypes` are replaced.
2) After it is changed, pls notify osd to change the name accordingly

nhtruong:
Closing since opensearchtypes have been replaced by generated types in 3.0.0

PR #433, opened 2023-03-14
[CCI][#206] renamed opensearchtypes CamelCase convention

BaurArabov:
### Description
renamed opensearchtypes to opensearchTypes due to the inconsistency in the naming convention of a module called "opensearchtypes".
This pull request updates the naming convention for the "opensearchtypes" module to follow the CamelCase naming convention used in OpenSearch Dashboards.

### Issues Resolved
https://github.com/opensearch-project/opensearch-js/issues/206


nhtruong:
Thanks @BaurArabov 
Would you mind updating the changelog and remove the `package-lock.json`?
We're currently using yarn as our package manager. So, having `package-lock.json` will be confusing and might cause issues in the future.

kavilla:
Can we have where we alias or map `opensearchtype` to `opensearchType` and make the before as deprecated. As is it would be a breaking change.

nhtruong:
@BaurArabov: @kavilla is right. This must be marked as deprecated first before we can include this change in the next Major update of the client.

nhtruong:
@BaurArabov The best way to do this is adding `opensearchTypes` with identical shape as `opensearchtypes`. Add a deprecation warning whenever `opensearchtypes` is used and direct the user to use `opensearchTypes` instead.

PR #442, opened 2023-03-18
[fix][CCI] Fixed typo in connection `_agent` 

### LINKED ISSUE #443: [BUG][CCI] Fixed typo in connection `_agent`
### What is the bug?

The declared variable [_agent](https://github.com/opensearch-project/opensearch-js/blob/main/lib/Connection.d.ts#L103) is local in the interface of the Connection class, here:

```
\\Connection.d.ts
export default class Connection {
...
  _agent: http.Agent | https.Agent | hpagent.HttpProxyAgent | hpagent.HttpsProxyAgent;
...
```

 The gist of the bug is that, it was discovered that in the class itself, the call goes to the agent variable, example:
```
if (typeof opts.agent === 'function') {
      this.agent = opts.agent(opts);
    } else if (opts.agent === false) {
      this.agent = undefined;
    }
```
Should be 
```
if (typeof opts.agent === 'function') {
      this._agent = opts.agent(opts);
    } else if (opts.agent === false) {
      this.a_gent = undefined;
    }
```

Naurabay:
### Description
Fixed typo in connection `_agent` 

### Issues Resolved
#443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added support for Amazon OpenSearch Serverless ([#356](https://github.com/opensearch-project/opensearch-js/issues/356))
 - Added required `array` argument when invoking `SerializationError` ([#419](https://github.com/opensearch-project/opensearch-js/pull/419))
 - Removed unnecessary `data` argument when invoking `OpenSearchClientError` ([#421](https://github.com/opensearch-project/opensearch-js/pull/421))
+- Fixed typos in `ConnectionPool` ([#427](https://github.com/opensearch-project/opensearch-js/pull/427))
 
 ### Dependencies
 - Bumps `xmlbuilder2` from 2.4.1 to 3.0.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Dependencies
 - Bumps `aws4` from 1.11.0 to 1.12.0
 - Bumps `minimist` from 1.2.6 to 1.2.8
-- Bumps `tsd` from 0.24.1 to 0.25.0
+- Bumps `tsd` from 0.24.1 to 0.27.0
 - Bumps `dezalgo` from 1.0.3 to 1.0.4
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added required `array` argument when invoking `SerializationError` ([#419](https://github.com/opensearch-project/opensearch-js/pull/419))
 - Removed unnecessary `data` argument when invoking `OpenSearchClientError` ([#421](https://github.com/opensearch-project/opensearch-js/pull/421))
 - Fixed typos in `ConnectionPool` ([#427](https://github.com/opensearch-project/opensearch-js/pull/427))
+- Fixed typo in connection `_agent` ([#443](https://github.com/opensearch-project/opensearch-js/issues/443)
```
dblock, line 50: You can skip CHANGELOGs for trivial things :)

lib/Connection.js
```diff
@@ -132,7 +132,7 @@ class Connection {
     const onAbort = () => {
       cleanListeners();
       request.once('error', () => {}); // we need to catch the request aborted error
-      debug('Request aborted', params);
```
dblock, line 135: Why?

dblock (commented):
I don't quite understand what the side effects of having `.agent` was. What was broken? Is there a test that you can write that shows broken behavior (that's fixed with this change)?

CHANGELOG.md
timursaurus: > I don't quite understand what the side effects of having `.agent` was. What was broken? Is there a test that you can write that shows broken behavior (that's fixed with this change)?

In the file [Connection.d.ts](https://github.com/opensearch-project/opensearch-js/blob/main/lib/Connection.d.ts#LL103C3-L103C87), the `agent` property is currently declared as `_agent`,  which has caused confusion.  It might be better to update the `Connection` interface to use `agent` instead, since it is not actually an internal property

@Naurabay 
```diff
export default class Connection {
  ...
  _status: string;
+ agent: http.Agent | https.Agent | hpagent.HttpProxyAgent | hpagent.HttpsProxyAgent;
- _agent: http.Agent | https.Agent | hpagent.HttpProxyAgent | hpagent.HttpsProxyAgent;
  constructor(opts?: ConnectionOptions) {
  ...
}
```

nhtruong:
@Naurabay are you still interested in solving this? I can take over if you want.

PR #430, opened 2023-03-12
[CCI] `BaseConnectionPool` missing implementation of `markAlive` and `markDead` methods

timursaurus:
### Description
Implement BaseConnectionPool `markDead` and `markAlive` methods
Fix unit test

### Issues Resolved
Closes #429 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #449, opened 2023-03-22
[CCI] `BaseConnectionPool` missing implementation of `markAlive` and `markDead` methods

### LINKED ISSUE #429: [BUG] [CCI] `BaseConnectionPool` empty methods
### What is the bug?
`BaseConnectionPool` methods [markAlive](https://github.com/opensearch-project/opensearch-js/blob/main/lib/pool/BaseConnectionPool.js#LL55C3-L55C15) and [markDead](https://github.com/opensearch-project/opensearch-js/blob/main/lib/pool/BaseConnectionPool.js#LL59C3-L59C11) are not implemented, but are called and passed connection argument, [example](https://github.com/opensearch-project/opensearch-js/blob/main/lib/pool/BaseConnectionPool.js#LL172C9-L172C40):
```js
if (connectionById) {
  debug(`The connection with id '${node.id}' is already present`);
  this.markAlive(connectionById);
```
They just return instance of the class. See: 
```js
 markAlive() {
    return this;
  }

  markDead() {
    return this;
  }
```

Two classes, [ConnectionPool](https://github.com/opensearch-project/opensearch-js/blob/main/lib/pool/ConnectionPool.js#LL63C3-L63C24) and [CloudConnectionPool](https://github.com/opensearch-project/opensearch-js/blob/main/lib/pool/CloudConnectionPool.js#LL34-L34C53), are extensions of the `BaseConnectionPool` class. While `ConnectionPool` implements the `markDead` and `markAlive` methods within the class, but `CloudConnectionPool` _does not_. This means that when using `CloudConnectionPool`, calling these methods has no effect.

timursaurus:
### Description

Implement `BaseConnectionPool` `markDead` and `markAlive` methods
Fix unit test

### Issues Resolved
Closes #429 

### Check List

  - [x] All tests pass
- [x] Changelog was updated.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -297,3 +297,26 @@ if (connection !== null) {
   client.connectionPool.markAlive(connection)
 }
 ```
+## Mark connection as `dead`
+```javascript
+console.log('Marking connection as dead:');
+
+var connection = client.connectionPool.getConnection()
+// or var connection = client.connectionPool.connections[0]
+
+if (connection !== null) {
+  client.connectionPool.markDead(connection)
+}
+```
+
+## Mark connection as `alive`
+```javascript
+console.log('Marking connection as alive:');
+
+var connection = client.connectionPool.getConnection()
+// or var connection = client.connectionPool.connections[0]
+
+if (connection !== null) {
+  client.connectionPool.markAlive(connection)
+}
+```
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -274,3 +274,26 @@ if (connection !== null) {
   client.connectionPool.markAlive(connection)
 }
 ```
+## Mark connection as `dead`
+```javascript
+console.log('Marking connection as dead:');
+
+var connection = client.connectionPool.getConnection()
+// or var connection = client.connectionPool.connections[0]
+
+if (connection !== null) {
+  client.connectionPool.markDead(connection)
+}
+```
+
+## Mark connection as `alive`
+```javascript
+console.log('Marking connection as alive:');
+
+var connection = client.connectionPool.getConnection()
+// or var connection = client.connectionPool.connections[0]
+
+if (connection !== null) {
+  client.connectionPool.markAlive(connection)
+}
+```
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -250,3 +250,27 @@ var response = await client.deleteAllPits();
 
 console.log(response.body);
 ```
+
+## Mark connection as `dead`
+```javascript
+console.log('Marking connection as dead:');
+
+var connection = client.connectionPool.getConnection()
+// or var connection = client.connectionPool.connections[0]
+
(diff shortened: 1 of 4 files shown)


test/unit/base-connection-pool.test.js
```diff
@@ -76,7 +76,18 @@ test('API', (t) => {
     let connection = pool.addConnection(href);
     t.same(pool.markDead(connection), pool);
     connection = pool.connections.find((c) => c.id === href);
-    t.equal(connection.status, Connection.statuses.ALIVE);
+    t.equal(connection.status, Connection.statuses.DEAD);
+    t.end();
+  });
+
+  t.test('call markDead twice', (t) => {
```
nhtruong, line 83: Why do you have to call markDead twice? Calling it once won't work?

timursaurus: > Why do you have to call markDead twice? Calling it once won't work?

https://github.com/opensearch-project/opensearch-js/pull/430#pullrequestreview-1336043602

USER_GUIDE.md
```diff
@@ -249,4 +249,28 @@ console.log('Deleting all PITs:');
 var response = await client.deleteAllPits();
 
 console.log(response.body);
+```
+
+## Mark connection as `dead`
```
dblock, line 254: I'd explain what the purpose of this call is, and ensure it fails when called twice.

timursaurus: By fail you mean a runtime error should be thrown when `markDead` is called on the same connection?

dblock: I think so. Unless there's a reason not to.

I still don't quite get how and when this API is used?

nhtruong: These methods are probably syntactical sugar for [set status](https://github.com/opensearch-project/opensearch-js/blob/790f610680450a2f8ea92e1985ac8afa7255d579/lib/Connection.js#L202-L205) method (There only 2 valid connection Statuses: DEAD or ALIVE). I still don't understand why it must throw an error when you set status to DEAD twice though. Wouldn't that also introduce new business logic that will cause trouble for users who are marking a connection as DEAD again?

nhtruong: Also since we have `set status` in Connection class, it makes more sense to move `markDead` and `markAlive` from ConnectionPool into Connection since you can change the status of the connection directly anyway. That is instead of `client.connectionPool.markAlive(connection)`, we will do `connection.markAlive()`. A lot more convenient and less confusing because it has nothing to do with ConnectionPool.

timursaurus: Sounds like an RFC to me
Should we implement this?

nhtruong: I think we should move it to `Connection` for the reason mentioned above

nhtruong:
@timursaurus do you wanna rebase this one so we can get it merged? Thank you :)

nhtruong:
Too stale

PR #451, opened 2023-03-24
[CCI] Type improvement and optional callback on `empty` method

### LINKED ISSUE #452: [BUG] [CCI] `BaseConnectionPool`'s `empty` method's callback is always required
### What is the bug?
[empty](https://github.com/opensearch-project/opensearch-js/blob/main/lib/pool/BaseConnectionPool.js#LL139C3-L139C20) method's `callback` argument of `BasePoolConnection` is always _required_, which could lead to runtime errors such as this:

Screenshot 2023-03-24 at 21 34 51

```js
const pool = new BaseConnectionPool({ Connection });
pool.addConnection('http://localhost:9200/');
pool.addConnection('http://localhost:9201/');
pool.empty()
```

Providing a callback should be optional

timursaurus:
### Description
Add more type safety
Make `empty` method's `callback` arg optional
Add/fix tests for listed above

### Issues Resolved
Closes #452 

### Check List

  - [x] All tests pass
- [x] Changelog was updated.


dblock:
Looks good. Needs USER_GUIDE and CHANGELOG updates pls.

timursaurus:
> Looks good. Needs USER_GUIDE and CHANGELOG updates pls.

On its way!

lib/pool/BaseConnectionPool.js
```diff
@@ -102,22 +103,30 @@ class BaseConnectionPool {
    */
   addConnection(opts) {
     if (Array.isArray(opts)) {
-      return opts.forEach((o) => this.addConnection(o));
-    }
+      opts.forEach((o) => this.addConnection(o));
```
nhtruong, line 106: Any reason why instead of returning this early, you decided to create an if/else block? Wouldn't that make the code harder to read?

timursaurus: I see. No reason actually, but let me tell you a story 
At first, i thought that returning a `forEach` was misleading  and removed it. End of story

Actually, i was playing around and it was odd to me that the function returns `void` and `Connection` which is not ideal, imo, but the in typescript definition it states that `addConnection` always returns `Connection`


I tried using Array.`map` and return array of connections,
```ts
addConnection(...): Connection | Connection[]
```
 but the tests failed and they were many, so I reverted everything back and the else block somehow stuck around :D

I'll remove the else block
But i'm moving the return statement to next line :P 
```ts
if (Array.isArray(opts)) {
      opts.forEach((o) => this.addConnection(o));
      return;
 }
```

USER_GUIDE.md
```diff
@@ -249,4 +249,19 @@ console.log('Deleting all PITs:');
 var response = await client.deleteAllPits();
 
 console.log(response.body);
+```
+
+## Empty all pool connections
```
dblock, line 254: Update TOC
Add a space below title
Capitalize Like Other Titles

USER_GUIDE.md
```diff
@@ -249,4 +249,19 @@ console.log('Deleting all PITs:');
...
+```javascript
+console.log('Emptying all pool connections:');
```
dblock, line 256: We should get rid of these `console.log`. It's self-explanatory what this code does, isn't it?

timursaurus: > We should get rid of these `console.log`. It's self-explanatory what this code does, isn't it?

Fair enough

timursaurus:
@dblock @nhtruong Any further comments or suggestions?

lib/pool/BaseConnectionPool.js
```diff
@@ -102,21 +103,29 @@ class BaseConnectionPool {
...
+      if (connectionById || connectionByUrl) {
+        throw new ConfigurationError(
+          `Connection with id '${connectionId || connectionUrl}' is already present`
+        );
+      }
     }
 
-    this.update([...this.connections, opts]);
+    const connection = this.createConnection(opts);
+    this.update([...this.connections, connection]);
```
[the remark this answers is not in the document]
nhtruong: Side Notes: This BaseConnectionPool class is very hard to read. We should refactor it in the future.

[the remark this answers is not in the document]
timursaurus: I see, let me give it a second look

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added Amazon OpenSearch Serverless in user_guide ([#372](https://github.com/opensearch-project/opensearch-js/issues/372))
 - Add guidelines on installing yarn, dependencies; instructions on running ESLint in developer_guide ([#439](https://github.com/opensearch-project/opensearch-js/issues/435))
 - Added pull_request_template ([440](https://github.com/opensearch-project/opensearch-js/pull/440))
+- Added test for creating a connection with an instance of Connection ([#451](https://github.com/opensearch-project/opensearch-js/pull/451))
```
dblock, line 13: No need for tests changelog

dblock (approved):
I would add a test that an empty pool can be emptied again, otherwise LGTM!

dblock (approved):

PR #454, opened 2023-03-24
[CCI] chore: cci workflow

timursaurus:
### Description

[CCI] 


nhtruong:
I'm not sure about this. This would be very convenient for sure but we don't let non-maintainers to add labels for security reasons and this will bypass that measure.

timursaurus:
> I'm not sure about this. This would be very convenient for sure but we don't let non-maintainers add labels for security reasons. And this will bypass that measure.

Totally expected, haha
Closing this

PR #456, opened 2023-03-24
[CCI] Update developerGuide regarding yarn troubleshoot steps

### LINKED ISSUE #455: [PROPOSAL] [CCI] Update developer's guide regarding Windows script execution policy
### What are you proposing?
Currently the Developer's Guide solution to Windows not letting you running scripts on client PCs is to set the policy to Unrestricted. Running the yarn install script should be possible even with the execution policy set to ```RemoteSigned```, and this will also prevent users from opening themselves up to potential malicious unsigned scripts that they can run.
Alternatively, we could add the line about reverting the security policy back to restricted after installing `yarn` package.

Wielmany:
### Description

Currently the Developer's Guide solution to Windows not letting you running scripts on client PCs is to set the policy to Unrestricted. Running the yarn install script should be possible even with the execution policy set to RemoteSigned, and this will also prevent users from opening themselves up to potential malicious unsigned scripts that they can run.
Besides this, by scoping the poilcy change to current user in the guide, there's no need to run PowerShell with administrator privileges (especially relevant when user doesn't have access to them)

### Issues Resolved

Closes #455 


Wielmany:
I think it might be worth it to just remove the part about setting 'Unrestricted' policy completely, everything seems to run fine with RemoteSigned anyway (but I left it for now in case someone has the problems somehow, as a last resort).
Should we remove it?

PR #532, opened 2023-06-06
bump `proxy` to 2.1.1 and use modular import

timursaurus:
### Description
Bumps `proxy` to `2.1.1`

### Issues Resolved
https://github.com/opensearch-project/opensearch-js/pull/531

### Check List

  - [ ] All tests pass
- [ ] Changelog was updated.


dblock:
Any external use changes? Let's make sure there are tests for this code path as part of this? And thanks!

nhtruong:
@timursaurus This will solve the failing Workflows we're seeing?

timursaurus:
> @timursaurus This will solve the failing Workflows we're seeing?

I'll look into it, seems like a cache problem with npm

timursaurus:
https://github.com/opensearch-project/opensearch-js/actions/runs/5245532512/jobs/9473116341?pr=532

`hpagent` requires Node 14 and higher
https://github.com/opensearch-project/opensearch-js/blob/main/.github/workflows/nodejs.yml#LL21C1-L21C47

wbeckler:
Just for context, bumping `proxy` solves a scanner-level CVE: https://github.com/advisories/GHSA-mj6p-3pc9-wf5m

I'm calling it "scanner-level" because the CVE is already mitigated by virtue of the way the library is used. But code scanners will hiccup on this until the version gets bumped.

nhtruong:
@timursaurus Did you mean to say drop 10.x only or 12.x too?
@wbeckler Node 10 came out in 2018. Do we expect the vast majority of our users to have moved on from this version of Node already? Do we a guideline for SLA in term of supporting legacy frameworks?

timursaurus:
> @timursaurus Did you mean to say drop 10.x only or 12.x too? 

 Yes, 12 too
 Some dependencies and AWS types require Node 14 and newer, which is the latest version of Node to reach End Of Life
 https://endoflife.date/nodejs

wbeckler:
In my experience, the bigger companies and applications can be stuck on node versions that are 5 or more years old, even if they are EOL. It's the sad reality. And it's a security nightmare.  I think the question of a node bump should get its own issue and discussion.

nhtruong:
@timursaurus we have some updates in the workflow regarding NPM that might have solved this as well. Please rebase and see if it works for ya.

PR #544, opened 2023-07-11
Add serialization and deserialization of numerals larger than `Number.MAX_SAFE_INTEGER`

AMoo-Miki:
### Description

In JavaScript, a `Number` is a 64-bit floating-point value which can store 16-ish digits. However, the serializer and deserializer will need to cater to numeric values generated by other languages which can have up to 19 digits. Native JSON parser and stringifier, incapable of handling the extra digits, corrupt the values, making them unusable.

To work around this limitation, with this change, the deserializer converts long sequences of digits into strings and marks them before applying the parser. During the parsing, string values that begin with the mark are converted to `BigInt` values. Similarly, during stringification, the serializer converts `BigInt` values to marked strings and when done, it replaces them with plain numerals.

### Issues Resolved

Related to https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4536

### Check List

  - [X] All tests pass
- [X] Changelog was updated.


kavilla (approved):

ananzh (approved):

AMoo-Miki:
Let's hold on merging this for a little while.

wbeckler:
We can do a patch release of the js client right now if that would help

test/unit/serializer.test.js
```diff
@@ -43,6 +43,30 @@ test('Basic', (t) => {
   t.same(s.deserialize(json), obj);
 });
 
+test('Long numerals', (t) => {
+  t.plan(6);
+  const s = new Serializer();
+  const longPositive = BigInt(Number.MAX_SAFE_INTEGER) * 2n; // eslint-disable-line no-undef
+  const longNegative = BigInt(Number.MIN_SAFE_INTEGER) * 2n; // eslint-disable-line no-undef
```
nhtruong, line 50: Can we also have a specific test case? Say "const number = 102931203123987", So when can avoid `a == a` test bug that's always true regardless of the logic.

nhtruong (changes requested):

nhtruong (approved):

lib/Serializer.js
```diff
@@ -44,11 +98,161 @@ class Serializer {
     };
   }
 
+  /* The characters with a highly unlikely chance of occurrence in strings, alone or in combination. */
+  _bigIntMarkChars = ['෴', '߷', '֍'];
+
+  /* Generates an array of all combinations of `_bigIntMarkChars` with the requested length. */
```
dblock, line 104: I know these are unlikely, but is there a more reliable way to have a BOM?

dblock (commented):
Did you write this code from scratch? :) 

- Is there really no BOM character that can never be present in the serialized data?
- Where is this serialization used? I think we should have at least one higher level test that shows it.

lib/Serializer.js
AMoo-Miki: It doesn't just use these guys. It creates a combination of 1 to Infinity of them. So even if these guys are all present, there is always a sequence that is long enough to not be present.

A future enhancement could dynamically generate a large collection based on ranges and experiment with them.

AMoo-Miki:
> Did you write this code from scratch? :)

Yes :)

>     * Is there really no BOM character that can never be present in the serialized data?

A string containing serialized JSON is still a JS string so all chars that can be used in JS strings can also be present in the serialized JSON.

PS, with the exception of U+2028 and U+2029, anything that goes into JSON, goes into JS, but then again, all of this serializing and deserializing is done in JS.

>     * Where is this serialization used? I think we should have at least one higher level test that shows it.

Transport.js uses it to produce the body of requests. It is also exported to be used by others. I will add an integration test.

nhtruong (approved):

lib/Serializer.js
dblock: I'm probably slow. These are all numbers. You get things like `෴18014398509481982`, why not `N:18014398509481982`?

PR #557, opened 2023-07-13
Make handling of long numerals an option that is disabled by default

AMoo-Miki:
### Description

This change adds a config param named `enableLongNumeralSupport` which is `false` by default and users who wish to deal with precise long numerals can enable it.

# This change needs to be added to the docs

### Check List

  - [X] All tests pass
- [ ] Changelog was updated.


dblock (changes requested):
What's the reason for this? What are the downsides of always serializing long numerals correctly?

What do you think about calling this `serializeLongNumerals`?

This PR needs an update to USERS_GUIDE, too, please.

AMoo-Miki:
> What's the reason for this? What are the downsides of always serializing long numerals correctly?

It is possible that users have adapted their logic to having imprecise numbers and are not ready to accept BigInts. One could argue that what we release in 2.3.0 was an enhancement but some could also argue that by successfully serializing BigInts (and not throwing) and deserializing long numerals into BigInts, we broke status quo and went against the universally known limitation of `JSON.parse` and `JSON.stringify` that people expect; one could argue that doing this by default is a breaking change.

> What do you think about calling this `serializeLongNumerals`?

We could surely name it `serializeLongNumerals` for those who will be using it by importing the `Serializer` but the `Client` needs a switch to use or not use this. Instead of the `Client` and `Helpers` each having conditions to use `serialize` or `serializeLongNumerals`, in this PR, I handle the condition in only one place which is inside `serialize`.

> This PR needs an update to USERS_GUIDE, too, please.

Done.

dblock (approved):

nhtruong:
looks like we need a 2.3.1 release?

wbeckler:
Can someone ELI5 why this is a breaking change?

dblock:
> Can someone ELI5 why this is a breaking change?

Unless I misunderstood something, we released 2.3.0 where the handling of long numerals is enabled by default, and 2.3.1 would disable it by default.

Issue #192, opened 2021-11-15
[BUG] DeserializationError: Unexpected token u in JSON at position 0


abagnani:
**Error:**
I found this bug while using the `client.helpers.bulk` helper function. When sending a batch of records into the function I was met with the following error: `DeserializationError: Unexpected token u in JSON at position 0`.

**Code:**
I was able to reproduce my error in the following script using a basic mock client and a simple batch of records as input.
```
#!/usr/bin/env node

const { Client } = require('@opensearch-project/opensearch');
const Mock = require('@elastic/elasticsearch-mock');

const mock = new Mock();

const client = new Client({
  node: 'http://localhost:9200',
  Connection: mock.getConnection(),
});

mock.add({
    method: 'POST',
    path: '/_bulk',
  }, () => ({
    errors: true,
    items: [
      { index: { _index: index, _id: '1', status: 404 } },
      { delete: { _index: index, _id: '2', status: 404 } },
      { index: { _index: index, _id: '3', status: 404 } },
    ],
  }));

const index = 'TEST_INDEX_NAME';

const testRecords = [
  { id: '1', key: 'test1' },
  { id: '2', key: 'test2', operation: 'remove' },
  { id: '3', key: 'test3' },
];

async function bulkTest () {
  const result = await client.helpers.bulk({
    datasource: testRecords,
    onDocument(doc) {
      if (doc.operation === 'remove') {
        return {
          delete: { _index: index, _id: doc.id },
        };
      }
      return {
        index: { _index: index, _id: doc.id },
      };
    },
    onDrop(doc) {
      console.log(doc);
    },
  });
  return result;
};

(async function() {
    const result = await bulkTest();
    console.log(result);
})();
```
The output of this script (`test.js`) is:
```
$ ./test.js
{
  status: 404,
  error: undefined,
  operation: { index: { _index: 'TEST_INDEX_NAME', _id: '1' } },
  document: { id: '1', key: 'test1' },
  retried: false
}
{
  status: 404,
  error: undefined,
  operation: { id: '1', key: 'test1' },
  document: null,
  retried: false
}
/@opensearch-project/opensearch/lib/Serializer.js:65
      throw new DeserializationError(err.message, json)
      ^
DeserializationError: Unexpected token u in JSON at position 0
    at Serializer.deserialize (/@opensearch-project/opensearch/lib/Serializer.js:65:13)
    at /@opensearch-project/opensearch/lib/Helpers.js:739:34
[...]

PR #620, opened 2023-10-03
Corrections to bulk.md

### LINKED ISSUE #616: Bulk API - Creating multiple documents
### What is the bug?
_Bulk method for creating multiple documents is not working properly._

_In the [Creating multiple documents](https://github.com/opensearch-project/opensearch-js/blob/main/guides/bulk.md#creating-multiple-documents) guide, there is a snippet of code how to create multiple documents._

`client.bulk({
  index: movies,
  body: [
    { create: { data: { title: 'Beauty and the Beast 2', year: 2030 } } },
    { create: { data: { title: 'Beauty and the Beast 3', year: 2031 } } },
    { create: { data: { title: 'Beauty and the Beast 4', year: 2049 } } },
    { create: { _index: books, data: { title: 'The Lion King 2', year: 1998 } } }
  ]
}).then((response) => {
  console.log(response);
});`

My Code Snippet:

`const bulk = async (index: string) => {
  try {
    const res = await OpenSearchClient.getClient().bulk({
      index,
      body: [
        {
          create: {
            data: {
              Id: '1',
              Title: 'Title 1',
              Language: 'en-US',
              ShortDocument: 'Short Doc 1',
              FullDocument: 'Full Doc 1',
            },
          },
        },
        {
          create: {
            data: {
              Id: '2',
              Title: 'Title 2',
              Language: 'en-US',
              ShortDocument: 'Short Doc 2',
              FullDocument: 'Full Doc 2',
            },
          },
        },
        {
          create: {
            data: {
              Id: '3',
              Title: 'Title 3',
              Language: 'en-US',
              ShortDocument: 'Short Doc 3',
              FullDocument: 'Full Doc 3',
            },
          },
        },
      ],
    });
    return res.body;
  } catch (error) {
    console.error(error);
    return { error };
  }
};`

Returns:

`ResponseError: illegal_argument_exception: [illegal_argument_exception] Reason: Malformed action/metadata line [1], expected a simple value for field [data] but found [START_OBJECT]
    at onBody (C:\Users\zoran\Documents\node_projects\opensearchdemo\node_modules\@opensearch-project\opensearch\lib\Transport.js:425:23)
    at IncomingMessage.onEnd (C:\Users\zoran\Documents\node_projects\opensearchdemo\node_modules\@opensearch-project\opensearch\lib\Transport.js:340:11)
    at IncomingMessage.emit (node:events:524:35)
    at endReadableNT (node:internal/streams/readable:1378:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  meta: {
    body: { error: [Object], status: 400 },
    statusCode: 400,
    headers: {
      'content-type': 'application/json; charset=UTF-8',
      'content-length': '339'
    },
    meta: {
      context: null,
      request: [Object],
      name: 'opensearch-js',
      connection: [Object],
      attempts: 0,
      aborted: false
    }
  }
}`

_Successfully create multiple documents._

### What is your host/environment?
_Windows 11 Pro, Version 22H2, OS Build 22621.2283. Node version v20.4.0. NPM version 9.7.2_

### Do you have any screenshots?
Screenshot

_Is the guide up to date or is it deprecated?_

AbhinavGarg90:
### Description

The description on how to use the API was incorrect. I had updated it with the correct implementation.

### Issues Resolved

#616 

### Check List

  - [ ] All tests pass
- [x] Changelog was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Deprecated
 ### Removed
 ### Fixed
+- Corrected documentation for `bulk.md` ([#620](https://github.com/opensearch-project/opensearch-js/pull/620))
 ### Security
 
 ## [2.3.0]
--- a/guides/bulk.md
+++ b/guides/bulk.md
@@ -52,18 +52,25 @@ client.bulk({
 ```
 As you can see, each bulk operation is comprised of two objects. The first object contains the operation type and the target document's `_index` and `_id`. The second object contains the document's data. As a result, the body of the request above contains six objects for three index actions.
 
-Alternatively, the `bulk` method can accept an array of hashes where each hash represents a single operation. The following code is equivalent to the previous example:
+Alternatively, the `bulk` method can accept an array of objects where objects are in pairs, except for delete which is singular. In each pair, the first item contains the action like `create` and, `_index` and `id` specify which index and document id the action has to be performed on. The second item in the pair is the document itself.
 
 ```javascript
 client.bulk({
-  body: [
-    { index: { _index: movies, _id: 1, data: { title: 'Beauty and the Beast', year: 1991 }  } },
-    { index: { _index: movies, _id: 2, data: { title: 'Beauty and the Beast - Live Action', year: 2017 } } },
-    { index: { _index: books, _id: 1, data: { title: 'The Lion King', year: 1994 } } }
-  ]
-}).then((response) => {
-  console.log(response);
-});
+    index: movies,
+    body: [
+      { create: { _id: 1} },
+      { title: 'Beauty and the Beast 1', year: 2050 },
+      { delete: { _id: 1 } },
+      { create: { _id: 2 } },
+      { title: 'Beauty and the Beast 2', year: 2051 },
+      { create: {} },
+      { title: 'Beauty and the Beast 2', year: 2051 },
+      { create: { _index: books } },
+      { title: '2012', year: 2012 },
+    ]
+  }).then((response) => {
+    console.log(response.body.items);
+  });
 ```
 
 We will use this format for the rest of the examples in this guide.
@@ -76,10 +83,14 @@ Similarly, instead of calling the `create` method for each document, you can use
 client.bulk({
   index: movies,
   body: [
-    { create: { data: { title: 'Beauty and the Beast 2', year: 2030 } } },
-    { create: { data: { title: 'Beauty and the Beast 3', year: 2031 } } },
-    { create: { data: { title: 'Beauty and the Beast 4', year: 2049 } } },
-    { create: { _index: books, data: { title: 'The Lion King 2', year: 1998 } } }
+    { create: {} },
+    { title: 'Beauty and the Beast 2', year: 2030 }, 
+    { create: {} },
+    { title: 'Beauty and the Beast 3', year: 2031 },
+    { create: {} },
+    { title: 'Beauty and the Beast 4', year: 2049 },
+    { create: { _index: books } },
+    { title: 'The Lion King 2', year: 1998 } 
   ]
 }).then((response) => {
   console.log(response);
@@ -92,8 +103,10 @@ Note that we specified only the `_index` for the last document in the request bo
 client.bulk({
   index: movies,
   body: [
-    { update: { _id: 1, data: { doc: { year: 1992 } } } },
-    { update: { _id: 2, data: { doc: { year: 2018 } } } }
+    { update: { _id: 1 } },
+    { doc: { year: 1992 } },
+    { update: { _id: 2 } },
+    { doc: { year: 2018 } }
   ]
 }).then((response) => {
   console.log(response);
@@ -122,9 +135,12 @@ You can mix and match the different operations in a single request. The followin
 client.bulk({
   index: movies,
   body: [
-    { create: { data: { title: 'Beauty and the Beast 5', year: 2050 } } },
-    { create: { data: { title: 'Beauty and the Beast 6', year: 2051 } } },
-    { update: { _id: 3, data: { doc: { year: 2052 } } } },
+    { create: {} },
+    { title: 'Beauty and the Beast 5', year: 2050 },
+    { create: {} },
+    { title: 'Beauty and the Beast 6', year: 2051 },
+    { update: { _id: 3 } },
+    { doc: { year: 2052 } },
     { delete: { _id: 4 } }
   ]
 }).then((response) => {
@@ -141,10 +157,14 @@ The following code shows how to look for errors in the response:
 client.bulk({
   index: movies,
   body: [
-    { create: { _id: 1, data: { title: 'Beauty and the Beast', year: 1991 } } },
-    { create: { _id: 2, data: { title: 'Beauty and the Beast 2', year: 2030 } } },
-    { create: { _id: 1, data: { title: 'Beauty and the Beast 3', year: 2031 } } }, // document already exists error
-    { create: { _id: 2, data: { title: 'Beauty and the Beast 4', year: 2049 } } }  // document already exists error
+    { create: { _id: 1 } },
+    { title: 'Beauty and the Beast', year: 1991 },
+    { create: { _id: 2} },
+    { title: 'Beauty and the Beast 2', year: 2030 },
+    { create: { _id: 1 } },
+    { title: 'Beauty and the Beast 3', year: 2031 }, // document already exists error
+    { create: { _id: 2 } },
+    { title: 'Beauty and the Beast 4', year: 2049 }  // document already exists error
   ]
 }).then((response) => {
   response.body.forEach((item) => {
@@ -162,5 +182,9 @@ client.bulk({
 To clean up the resources created in this guide, delete the `movies` and `books` indices:
 
 ```javascript
-client.indices.delete({ index: [movies, books] });
+client.indices.delete({
+    index: [movies, books]
+}).then((response) => {
+    console.log(response);
+});
 ```


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Make handling of long numerals an option that is disabled by default ([#557](https://github.com/opensearch-project/opensearch-js/pull/557))
 - Add `enableLongNumeralSupport` to the type definition for `ClientOptions` ([#598](https://github.com/opensearch-project/opensearch-js/pull/598))
+- Corrected documentation for bulk.md ([#XXX])
```
dblock, line 25: This should go into fixes. Please update the XXX to the PR number.

dblock (commented):
Do you think you could add a working sample into `samples` like https://github.com/opensearch-project/opensearch-py/tree/main/samples as part of this PR?

nhtruong:
> Do you think you could add a working sample into `samples` like https://github.com/opensearch-project/opensearch-py/tree/main/samples as part of this PR?

That's entirely different from `guides`. The equivalent on Python client is this one https://github.com/opensearch-project/opensearch-py/blob/main/guides/bulk.md

nhtruong:
> > > Do you think you could add a working sample into `samples` like https://github.com/opensearch-project/opensearch-py/tree/main/samples as part of this PR?
> > 
> > 
> > That's entirely different from `guides`. The equivalent on Python client is this one https://github.com/opensearch-project/opensearch-py/blob/main/guides/bulk.md
> 
> Should I try to update that bulk.md file to be closer in line to [the ruby version](https://github.com/opensearch-project/opensearch-ruby/blob/main/guides/bulk.md) ?

You're welcome to. Or you can also create an issue in the Python Repo.

guides/bulk.md
```diff
@@ -52,18 +52,25 @@ client.bulk({
 ```
 As you can see, each bulk operation is comprised of two objects. The first object contains the operation type and the target document's `_index` and `_id`. The second object contains the document's data. As a result, the body of the request above contains six objects for three index actions.
 
-Alternatively, the `bulk` method can accept an array of hashes where each hash represents a single operation. The following code is equivalent to the previous example:
+Alternatively, the `bulk` method can accept an array of hashes where hashes are in pairs, except for delete which is singular. In each pair, the first item contains the action like `create` and, `_index` and `id` specify which index and document id the action has to be performed on. The second item in the pair is the document itself.
```
nhtruong, line 55: `hash`, a very common data structure in Ruby, is actually referred to as an Object (as in JSON Object) in JavaScript.

guides/bulk.md
```diff
@@ -52,18 +52,25 @@ client.bulk({
...
+      { title: 'Beauty and the Beast 2', year: 2051 },
+      { create: { _index: books } },
+      { title: '2012', year: 2012 },
+    ]
+  }).then((response) => {
+    console.log(response.body.items);
+  });
 ```
 
 We will use this format for the rest of the examples in this guide.
```
nhtruong, line 76: Looks like this part is not applicable to JavaScript. There's only way to format the body for `client.bulk` in JS unlike in Ruby.

nhtruong: As in we can remove line 55 to 76

nhtruong: Look like we have `client.helpers.bulk` that allows another format if you wanna look into that.

guides/bulk.md
AbhinavGarg90: Will correct that now, I was apprehensive of it when I saw it earlier.

guides/bulk.md
AbhinavGarg90: > As in we can remove line 55 to 76

I don't follow exactly. Is that not how client.bulk works? Because I have used that same method across the document.

dblock: I think you should write a working sample, and include it along with these changes so we can be sure @AbhinavGarg90.

nhtruong: @dblock The code snippets  in each guide are supposed to be working examples that can be run from start to finish, in order, without throwing errors. It's the reason @AbhinavGarg90 is fixing it right now because the current snippets throw errors.


@AbhinavGarg90 OpenSearch only accepts 1 body format for the bulk action which is an array of operation/document pair like so:
```ruby
 [
    { create: { _index: movies, _id: 1 } },
    { title: 'Beauty and the Beast', year: 1991 },
    { delete: { _index: movies, _id: 1 } },
    { index:  { _index: books, _id: 2 } },
    { title: 'The Lion King', year: 1994 }
 ]
 ```
 
 However, the implementation of this action on Ruby also accepts another format that's simpler:
 ```ruby
[
    { create: { _index: movies, _id: 1, data: { title: 'Beauty and the Beast', year: 1991 }  } },
    { delete: { _index: movies, _id: 1},
    { index:  { _index: books, _id: 1, data: { title: 'The Lion King', year: 1994 } } }
]
```
This format is then translated to the former one before sending it is serialized and sent to the server.
`client.bulk` in javascript doesn't support this format. However looks like `client.helpers.bulk` also provides similar syntactical sugar. That's why I think you should look into that and add it to this guide as well, because users looking to use `client.bulk` will likely wanna know about `client.helpers.bulk` as well.

dblock: > The code snippets in each guide are supposed to be working examples 

I'm just saying that we need to commit the working samples and run them if we want them to be ... working.

AbhinavGarg90: > Look like we have `client.helpers.bulk` that allows another format if you wanna look into that.

I can work on adding this, can you explain a bit on how this function is supposed to be called? I looked through the code, and I'm not completely sure.

dblock: I don't have good sample code, but https://opensearch.org/docs/2.2/clients/javascript/helpers/ seems to be quite thorough, and I found a few people reporting bugs with otherwise working examples, e.g [here](https://github.com/opensearch-project/opensearch-js/issues/192).

AbhinavGarg90: Great, I'll look through this and add it to the pull request. Once I'm done with that I'll try to look into the other issue as well.

AbhinavGarg90:
```js
const { Client } = require('@opensearch-project/opensearch')
const documents = require('./docs.json')

const client = new Client({ ... })

const result = await client.helpers.bulk({
  datasource: documents,
  onDocument (doc) {
    return {
      index: { _index: 'example-index' }
    }
  }
})

console.log(result)
```

Looking at the documentation, it doesn't specify the shape of the input document. I assume it has key and value, but I cannot seem to locate any documentation with further detail about this.

dblock: You mean https://opensearch.org/docs/latest/api-reference/document-apis/bulk/? It's pretty complete (bulk API accepts line-delimited JSON).

dblock:
> sample

I think we need this to convince ourselves that the documentation is now correct.

guides/bulk.md
AbhinavGarg90: > You mean https://opensearch.org/docs/latest/api-reference/document-apis/bulk/? It's pretty complete (bulk API accepts line-delimited JSON).

Thank you for pointing out where this is. Should I link this in bulk.md as well?

AbhinavGarg90:
Hey @dblock I realized that I had created this PR from main on my fork. I want to move it to a feature branch on my fork. How can I do so?

I think syncing main from my fork accidently closed this PR. How can I reopen it?

guides/bulk.md
AbhinavGarg90: How does line-delimited JSON work? I tried running a sample using the available documentation, it doesn't seem to be working.

Here are the code snippets I am working with:

```js
const { Client } = require('@opensearch-project/opensearch');
const documents = require('./docs.json')

try {
  const client = new Client({
    node: 'https://admin:admin@localhost:9200',
    ssl: { rejectUnauthorized: false }
  });
  client.helpers.bulk({
    datasource: documents,
    onDocument (doc) {
      return {
        index: { _index: 'example-index' }
      }
    }
  }).then((response) => {
    console.log(response)
  }).catch((error) => {
    console.error(error);
  })
} catch(e) {
  console.log("error", e)
}

```
docs.json
```json
{ "delete": { "_index": "movies", "_id": "tt2229499" } }
{ "index": { "_index": "movies", "_id": "tt1979320" } }
{ "title": "Rush", "year": 2013 }
{ "create": { "_index": "movies", "_id": "tt1392214" } }
{ "title": "Prisoners", "year": 2013 }
{ "update": { "_index": "movies", "_id": "tt0816711" } }
{ "doc" : { "title": "World War Z" } }
```

error:
```sh
Unexpected non-whitespace character after JSON at position 57
```

nhtruong: @AbhinavGarg90 Sorry for the very late response. I've been out of commission for the last few weeks.

The datasource is an iterable and the simplest iterable in this case is an array of objects, where each object is an OpenSearch doc:
```javascript
  const docs = [{ title: 'Beauty and the Beast 1', year: 2050 },
                { title: 'Beauty and the Beast 2', year: 2051 }]
```
Then you can pass it to the helper function:
```javascript
  let result = await client.helpers.bulk({ datasource: docs,
    onDocument(doc) {
      return { index: { _index: movies} }
    }
  });

  console.log(result);
```
And you should get this in the console:
```
{
  total: 2,
  failed: 0,
  retry: 0,
  successful: 2,
  noop: 0,
  time: 744,
  bytes: 150,
  aborted: false
}
```

Basically the `client.helpers.bulk` method, on top of implementing some batching logic, also acts as syntactical sugar for `client.bulk`. That is, the user only have to provide the documents in their original forms (an array of objects in the above example) and the action to apply (index in this example) to apply to the documents.

nhtruong:
@AbhinavGarg90 You'd have to recreate the PR because the source and target branches are locked. However, for now, you can simply rebase your main branch onto the `upstream/main` branch, then force push, if you want the updates from `upstream/main`.

nhtruong:
The guide has been reworked.

PR #665, opened 2023-12-09
[Update] delete index guide with ignore_unavailable

### LINKED ISSUE #658: [BUG] Documented Parameter for Delete-Index API Call is Unsupported
### What is the bug?
Within the Index Lifecycle guide, [the "Delete an Index" example](https://github.com/opensearch-project/opensearch-js/blob/2.4/guides/index_lifecycle.md#delete-an-index), demonstrates that an `ignore: 404` parameter should be passed in order to prevent an error from being returned if one or more indices do not exist.

When using this parameter, however, an error is thrown with the message: `request [/{index-name-here}] contains unrecognized parameter: [ignore]` (where `{index-name-here}` is the name of the index that was identified for deletion. A similar error occurs if you pass an array (as in the example) of index names.

In looking at [the code for the delete-index method](https://github.com/opensearch-project/opensearch-js/blob/2.4/api/api/indices.js) (line 444 begins the method definition), there indeed seems to be no such parameter accepted. There is, however, a Boolean `ignore_unavailable` parameter which seems to do the job.

Call the delete-index API method, passing the name of an index that does not exist and passing `ignore: 404` as in the Index Lifecycle example.

According to the Index Lifecycle guide, no error should be thrown.

### What is your host/environment?
Host environment is an AWS Lambda function, so some version of Linux running our Node.js lambda (which exercises the `@opensearch-project/opensearch` API.

### Proposed Resolution
I don't quite care which approach is taken, but one of the following should be taken with preference given to whichever best aligns with the mission and/or future direction of this project.

1. Update the _documentation_ to reflect the correct parameter to pass that will cause the index-delete API call to _not_ throw an error when an index does not exist.
2. Update the _code_ (the `IndicesApi.prototype.delete`  function) to operate as the documentation prescribes.

tejeshreddy:
### Description

Update **Delete an Index** in index_lifecycle.md with `ignore_unavailable: true`

### Issues Resolved

#658

### Check List

- [X]  Changelog was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
-- Updated `guides/index_lifecycle.md` to provide example of `ignore_unavailable: true` while deleting indices.
+- Updated `guides/index_lifecycle.md` to provide example of `ignore_unavailable: true` while deleting indices. ([665](https://github.com/opensearch-project/opensearch-js/pull/665))
 ### Dependencies
 - Bumps `@types/node` from 20.9.0 to 20.10.3
 - Bumps `eslint` from 8.54.0 to 8.55.0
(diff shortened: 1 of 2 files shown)


nhtruong (approved):

nhtruong:
Thanks for the work. :D

Github tip for next time: In the description if you say "closes https://github.com/opensearch-project/opensearch-js/issues/658", that issue will also be closed automatically once this PR is merged.

PR #669, opened 2023-12-13
Pass in initialAdminPassword

derek-ho:
### Description

Due to recent changes in the security plugin, the install demo configuration script now requires an initial admin password to be set. This sets it to some random variable so that the CI continues to work.

### Issues Resolved

### Check List

  - [ ] All tests pass
- [ ] Changelog was updated.

### DIFF
--- a/.ci/opensearch/Dockerfile.opensearch
+++ b/.ci/opensearch/Dockerfile.opensearch
@@ -9,6 +9,6 @@ ARG SECURE_INTEGRATION
 HEALTHCHECK --start-period=20s --interval=5s --retries=2 --timeout=1s \
   CMD if [ "$SECURE_INTEGRATION" != "true" ]; \
   then curl --fail localhost:9200/_cat/health; \
-  else curl --fail -k https:/localhost:9200/_cat/health -u admin:admin; fi
+  else curl --fail -k https:/localhost:9200/_cat/health -u admin:myStrongPassword123!; fi
 
 RUN if [ "$SECURE_INTEGRATION" != "true" ] ; then $opensearch_path/bin/opensearch-plugin remove opensearch-security; fi
--- a/.ci/opensearch/docker-compose.yml
+++ b/.ci/opensearch/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - discovery.type=single-node
       - bootstrap.memory_lock=true
       - SECURE_INTEGRATION=${SECURE_INTEGRATION:-false}
-      - initialAdminPassword=myStrongPassword123!
+      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!
     ports:
       - '9200:9200'
     user: opensearch
--- a/.ci/opensearch/docker-compose.yml
+++ b/.ci/opensearch/docker-compose.yml
@@ -13,6 +13,7 @@ services:
       - discovery.type=single-node
       - bootstrap.memory_lock=true
       - SECURE_INTEGRATION=${SECURE_INTEGRATION:-false}
+      - initialAdminPassword=myStrongPassword123!
     ports:
       - '9200:9200'
     user: opensearch
--- a/guides/msearch.md
+++ b/guides/msearch.md
@@ -8,7 +8,7 @@ OpenSearch's Multi-Search (`msearch`) API allows you to execute multiple search 
 const host = "localhost";
 const protocol = "https";
 const port = 9200;
-const auth = "admin:admin";
+const auth = "admin:${initialAdminPassword}";
 const ca_certs_path = "/full/path/to/root-ca.pem";
 const { Client } = require("@opensearch-project/opensearch");
 const fs = require("fs");
--- a/guides/msearch.md
+++ b/guides/msearch.md
@@ -8,7 +8,7 @@ OpenSearch's Multi-Search (`msearch`) API allows you to execute multiple search 
 const host = "localhost";
 const protocol = "https";
 const port = 9200;
-const auth = `admin:${initialAdminPassword}`;
+const auth = `admin:<admin password>`;
 const ca_certs_path = "/full/path/to/root-ca.pem";
 const { Client } = require("@opensearch-project/opensearch");
 const fs = require("fs");
--- a/guides/msearch.md
+++ b/guides/msearch.md
@@ -8,7 +8,7 @@ OpenSearch's Multi-Search (`msearch`) API allows you to execute multiple search 
 const host = "localhost";
 const protocol = "https";
 const port = 9200;
-const auth = `admin:<admin password>`;
+const auth = `admin:myStrongPassword123!`;
 const ca_certs_path = "/full/path/to/root-ca.pem";
 const { Client } = require("@opensearch-project/opensearch");
 const fs = require("fs");
--- a/guides/msearch.md
+++ b/guides/msearch.md
@@ -8,7 +8,7 @@ OpenSearch's Multi-Search (`msearch`) API allows you to execute multiple search 
 const host = "localhost";
 const protocol = "https";
 const port = 9200;
-const auth = "admin:${initialAdminPassword}";
+const auth = `admin:${initialAdminPassword}`;
 const ca_certs_path = "/full/path/to/root-ca.pem";
 const { Client } = require("@opensearch-project/opensearch");
 const fs = require("fs");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Bumps `@aws-sdk/types` from 3.451.0 to 3.468.0
 - Bumps `prettier` from 3.1.0 to 3.1.1
 ### Changed
-- Pass in initialAdminPassword to docker and update instances of admin:admin in READMEs. ([669](https://github.com/opensearch-project/opensearch-js/pull/669))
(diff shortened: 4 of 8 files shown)


dblock (changes requested):
To fix CI don't you just need to say `initialAdminPassword=admin` in `docker-compose.yml`? 

I think the documentation should be left alone. Or we should explain how to obtain that initial password.

derek-ho:
> To fix CI don't you just need to say `initialAdminPassword=admin` in `docker-compose.yml`?
> 
> I think the documentation should be left alone. Or we should explain how to obtain that initial password.

Since "admin" is considered a weak password, this wouldn't be possible. There is weak validation in place, and it needs a strong password. Password validation is being done via this library: https://github.com/dropbox/zxcvbn.

dblock:
> > To fix CI don't you just need to say `initialAdminPassword=admin` in `docker-compose.yml`?
> > I think the documentation should be left alone. Or we should explain how to obtain that initial password.
> 
> Since "admin" is considered a weak password, this wouldn't be possible. There is weak validation in place, and it needs a strong password. Password validation is being done via this library: https://github.com/dropbox/zxcvbn. It is possible to pass in a weak password if running the install-demo-configuraiton script by yourself, but if you are using the distributions we don't currently support it.

I see. I think we should hardcode a non-readable password in our docker setup.

I would say `admin:<initial password>` rather than `${initialAdminPassword}` that implies some kind of variable.

derek-ho:
> > > To fix CI don't you just need to say `initialAdminPassword=admin` in `docker-compose.yml`?
> > > I think the documentation should be left alone. Or we should explain how to obtain that initial password.
> > 
> > 
> > Since "admin" is considered a weak password, this wouldn't be possible. There is weak validation in place, and it needs a strong password. Password validation is being done via this library: https://github.com/dropbox/zxcvbn. It is possible to pass in a weak password if running the install-demo-configuraiton script by yourself, but if you are using the distributions we don't currently support it.
> 
> I see. I think we should hardcode a non-readable password in our docker setup.

I didn't get this point - do you just mean something that is random/doesn't contain whole words?

> I would say `admin:<initial password>` rather than `${initialAdminPassword}` that implies some kind of variable.

Done

guides/msearch.md
```diff
@@ -8,7 +8,7 @@ OpenSearch's Multi-Search (`msearch`) API allows you to execute multiple search
 const host = "localhost";
 const protocol = "https";
 const port = 9200;
-const auth = "admin:admin";
```
nhtruong, line 11: This code block is supposed to be able to executed as is. This change will cause an error to throw right?

nhtruong: Same for similar changes in the `./guides` folder.

derek-ho: Yes most likely it will cause an error, but we are intending users to set admin passwords now. I see a similar behavior in the line below:
```
const ca_certs_path = "/full/path/to/root-ca.pem";
```
Which seems to expect users to change this area?

nhtruong: Can we change it to `admin:myStrongPassword123!` since you already have that set in the dockerfile? The goal is lower the barrier of entry for the users so that they have to make as few changes as possible :)

derek-ho: Done in: 47b573e

derek-ho:
@dblock @nhtruong can you take another look at this PR? I'm wondering as to why the CI doesn't break, since `admin:myStrongPassword123!` shouldn't work on secure clusters until 2.12.0 is released. Let me know what you folks think is an appropriate path forward for this PR, thanks! CIs of other PRs are also passing, so I am suspecting we may not even need this PR - I don't think the healthcheck command is working correctly.

DarshitChanpura:
@dblock @nhtruong Can we get more reviews on this PR?

dblock:
@DarshitChanpura take a look at @derek-ho's comment above. 

1. Why is this not breaking?
2. Do we need this change before 2.12?

derek-ho:
Putting this PR in draft until 2.12.0 is released, since it looks like the workflows are only using released docker versions from opensearchproject. I am still confused why changing the healthcheck command to non - "admin" password is still passing, since released versions up until now should not take up the new password. Maybe something to look into after release.

DarshitChanpura:
@dblock, Derek and I are looking to get feedback from maintainers here to understand whether CI is run against 2.12 and if so is there any codepath that we should address. Right now, based on our understanding, we suspect that the health checks might not be working, and we would like the feedback on the path forward for this PR.

We are not familiar with this code-base and initially raised this PR (and others across different repositories) to help with the changes needed as part of larger effort: https://github.com/opensearch-project/security/issues/3624 and are seeking guidance from maintainers on the path forward.

nhtruong:
@DarshitChanpura 

0) CI is running against 2.x and main: https://github.com/opensearch-project/opensearch-js/blob/main/.github/workflows/integration-unreleased.yml#L24-L25
1) Shouldn't the password use in this PR work against < 2.12 too?

.ci/opensearch/Dockerfile.opensearch
```diff
@@ -9,6 +9,6 @@ ARG SECURE_INTEGRATION
 HEALTHCHECK --start-period=20s --interval=5s --retries=2 --timeout=1s \
   CMD if [ "$SECURE_INTEGRATION" != "true" ]; \
   then curl --fail localhost:9200/_cat/health; \
-  else curl --fail -k https:/localhost:9200/_cat/health -u admin:admin; fi
+  else curl --fail -k https:/localhost:9200/_cat/health -u admin:myStrongPassword123!; fi
```
DarshitChanpura, line 12: Should this change be version based? `admin` for <2.12 and `myStrongPassword123!` for >=2.12.

DarshitChanpura (commented):
> 0. CI is running against 2.x and main: https://github.com/opensearch-project/opensearch-js/blob/main/.github/workflows/integration-unreleased.yml#L24-L25

If I am reading the workflow right, the workflow uses min OpenSearch distribution in which case security will not be installed and no admin password is required. https://github.com/opensearch-project/opensearch-js/blob/18389de443a7b5d997a8052f7136e17b3f3fb067/.github/workflows/integration-unreleased.yml#L61


> 1. Shouldn't the password used in this PR work against < 2.12 too?

The password change is only for versions 2.12 and later. For versions <2.12 the password is still `admin`.

nhtruong:
> If I am reading the workflow right, the workflow uses min OpenSearch distribution in which case security will not be installed (as it doesn't exist as part of min distribution) and no admin password is required.


The CI uses 2 different configuration for security: On and Off https://github.com/opensearch-project/opensearch-js/blob/main/.ci/opensearch/Dockerfile.opensearch#L14

I believe we need to turn it on for that workflow like this https://github.com/opensearch-project/opensearch-js/blob/main/.github/workflows/integration.yml#L80

DarshitChanpura:
> The CI uses 2 different configuration for security: On and Off https://github.com/opensearch-project/opensearch-js/blob/main/.ci/opensearch/Dockerfile.opensearch#L14


@nhtruong Correct, but that comes into play only if a full distribution is used. A min distribution of opensearch does not contain opensearch-security, and this is what I see as being used here: https://github.com/opensearch-project/opensearch-js/blob/18389de443a7b5d997a8052f7136e17b3f3fb067/.github/workflows/integration-unreleased.yml#L61

and in that case the flag will not matter.

DarshitChanpura:
For released versions workflow `integration.yml` the changes introduced in this PR (with a modification I suggested [above](https://github.com/opensearch-project/opensearch-js/pull/669#discussion_r1463830686)) will be required once 2.12 is released and added to the workflow matrix

DarshitChanpura:
> We can make the tests using security plugin to pull password from an env variable and have it fallback to admin if the variable does not exist.

That fallback simply won't work as there is a strong password check in place and `admin` is not a strong password.

The proposed approach here is to conditionally use the custom (e.g. `myStrongPassword123!`) / default (`admin`) password based on the version being tested.

nhtruong:
I meant setting the client's env var to the strong password when we use 2.12 and up. But either works :)

nhtruong:
@DarshitChanpura for sure. Lemme find a good stopping point for my current project before switching context. I will be able get on this tomorrow or Thu.

DarshitChanpura:
Ty so much @nhtruong !!

nhtruong:
I've started taking a crack at this. Will create a new PR but leaving this open for discussion.

DarshitChanpura:
>I've started taking a crack at this. Will create a new PR but leaving this open for discussion.

Thank you @nhtruong. Btw, as a maintainer you should have access to pushing changes directly on this PR.
You can simply checkout this PR: `gh pr checkout 669` and starting commiting/pushing the changes.

nhtruong:
Cant push to someone else's protected branch (main) unfortunately. Imma create a new draft to avoid spamming yall with notifications anyway.

.ci/opensearch/docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
       - discovery.type=single-node
       - bootstrap.memory_lock=true
       - SECURE_INTEGRATION=${SECURE_INTEGRATION:-false}
+      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!
```
nhtruong, line 16: @DarshitChanpura with OS 2.12 and Up, would this be enough to set the admin password? Does the plugin just grab what stored in `OPENSEARCH_INITIAL_ADMIN_PASSWORD`, check if it's strong enough, and then assign it as password for `admin` user?

nhtruong: I also saw this in the security plugin repo: https://github.com/opensearch-project/security/pull/3843/files
Does that mean that if we set `initialAdminPassword` to `myStrongPassword123!`, it will also work for older version of OS?

DarshitChanpura:
Closing this PR in lieu of #707

Issue #670, opened 2023-12-14
[BUG] Security namespace is missing from the Client type declaration


stowns:
### What is the bug?
_A clear and concise description of the bug._
It appears that `security` is missing from the [Client type declaration](https://github.com/opensearch-project/opensearch-js/blob/main/index.d.ts#L144) which causes "Property 'security' does not exist on type 'Client'" in IDE's.

### How can one reproduce the bug?
_Steps to reproduce the behavior._
```
import { Client  } from '@opensearch-project/opensearch';
const client = new Client(...)
client.security
```

### What is the expected behavior?
_A clear and concise description of what you expected to happen._
The `security` namespace exists on the client.

### What is your host/environment?
_Operating system, version._
MacOS 13.6.2

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._
@opensearch-project/opensearch: 2.4.0

The tests [added in the commit](https://github.com/opensearch-project/opensearch-js/pull/606/commits/1037df42a4e4f077d77869e9b0709a00f399d289) are in javascript. I believe this is why the issue wasn't observed

nhtruong:
It's already been added to https://github.com/opensearch-project/opensearch-js/blob/main/api/new.d.ts#L1966
I'm not sure why we have 2 places to declare types for API methods 😅

GTRanger:
The `new.d.ts` export is not what is used by default with the import statement `import { Client  } from '@opensearch-project/opensearch';`. Rather, it still points to the definitions in `index.d.ts`. The default exports need to be updated to use the new type definitions.

nhtruong:
@GTRanger Did this [PR](https://github.com/opensearch-project/opensearch-js/pull/674/files) resolve the issue?

nhtruong:
This has been resolved by generated types from 3.0.0

PR #674, opened 2023-12-18
Add new types to package exports definition for ESM support

### LINKED ISSUE #204: Redirect types in package.json to api/new.d.ts
Currently, in package.json, types is still using index.d.ts. 
```
"name": "@opensearch-project/opensearch",
  "description": "The official OpenSearch client for Node.js",
  "main": "index.js",
  "types": "index.d.ts",
  "exports": {
    ".": {
      "require": "./index.js",
      "import": "./index.mjs"
    },
    "./": "./"
  },

...
```
However, the api types have been updated and the new types are api/new.d.ts. Therefore, we need either change `types` in package.json or copy api/new.d.ts to index.d.ts

perttukarna:
### Description

As of today, it's not possible to resolve the new typings with a TS project using ESM. With these changes, you can import types in ESM project, e.g.:

```typescript
import type { SearchHit, SearchResponse } from '@opensearch-project/opensearch/api/types'
```

This is also backwards compatible with projects converting from TS + CommonJS to TS + ESM.

### Issues Resolved

Closes #204 

### Check List

  - [ ] All tests pass
- [x] Changelog was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Deprecated
 ### Removed
 ### Fixed
+- Add new types to `package.json` exports configuration for ESM support ([#674](https://github.com/opensearch-project/opensearch-js/pull/674))
 ### Security
 
 ## [2.5.0]
--- a/package.json
+++ b/package.json
@@ -9,6 +9,12 @@
       "types": "./index.d.ts",
       "import": "./index.mjs"
     },
+    "./api/new": {
+      "types": "./api/new.d.ts"
+    },
+    "./api/types": {
+      "types": "./api/types.d.ts"
+    },
     "./aws": "./lib/aws/index.js",
     "./*": "./*"
   },


kavilla:
@AMoo-Miki

dblock:
Is there a way to add a test for this so we don't break it in the future?

perttukarna:
> Is there a way to add a test for this so we don't break it in the future?

Unfortunately I'm not sure if there is an easy way to test it - it probably requires setting up a child project with ESM and type checking.

However, after having another look at the type definitions, I noticed the `./api/types` -module is already exported from the root typings as [opensearchtypes namespace export](https://github.com/opensearch-project/opensearch-js/blob/main/index.d.ts#L5623), similarly as `./api/requestParams`. Should we export just the `./api/new` as explicit ESM (type) export, and request users to import the types from `./api/types` via the root package namespace export, such as:

```typescript
import { ApiResponse, opensearchtypes } from '@opensearch-project/opensearch'
import { Client } from '@opensearch-project/opensearch/api/new'

const client: Client
const response: ApiResponse<opensearchtypes.SearchResponse<TDocument>> = await client.search({ ... })
```

Technically we could export the new types via the root typings as well, but if they're still partially work in progress and opt-in, I'd keep them separate.

perttukarna:
@dblock @kavilla @AMoo-Miki not trying to rush anything, but happy to improve upon your request.

I'd suggest we change this PR to export only the new types 

```json
    "./api/new": {
      "types": "./api/new.d.ts"
    }
```

What do you think?

nhtruong:
@perttukarna Is this related to https://github.com/opensearch-project/opensearch-js/issues/670#issuecomment-1864866676 ?

perttukarna:
@nhtruong yes and no. 😄 This won't solve the problem of having two different type definitions for the client, but merely makes the new type definitions available for certain project setups (TS + ESM in particular).

perttukarna:
@dblock @kavilla @AMoo-Miki can you take a look, please?

dblock (approved):
AFAUK this looks correct, but 🤷‍♂️

Issue #677, opened 2023-12-20
[FEATURE] Expiration buffer for credentials


ozonep:
### Is your feature request related to a problem?
If job takes long time to complete (more than credentials expiry time), OpenSearch client may have issues renewing credentials upon their expiry, so requests start to fail in case AWS SDK v3 is used.
We found this to be an issue once we migrated to AWS SDK v3. We have jobs that may take 40 hours to complete, and requests almost always start to fail exactly on 1 hour mark (that is our credentials expiry time, max value allowed by AWS). Sometimes this happens after 2 hours, sometimes after 3 hours, but always does.

### What solution would you like?
Add expiration buffer value that is at least a value of `requestTimeout`, so in `awsSigner` module expiration check would be:

```javascript
const EXPIRATION_BUFFER_MS = 60_000 // or pass value of `requestTimeout`
if (currentCredentials.expiration?.getTime() - Date.now() < EXPIRATION_BUFFER_MS) {
   expired = true;
}
```

Setting expiration buffer to 5 minutes fixed our issue (we have `requestTimeout` of 1 minute, but we used 5 minutes as buffer just to be on the safe side).
But IMO value should be at least the value of `requestTimeout` so we don't start request with credentials that will expire sooner than request may time out.

I can create PR with this feature.

### What alternatives have you considered?
Currently I see 3 ways of setting this buffer value:
- Have it as a static value (1 minute? 30 seconds?)
- Allow passing this value to the client as an option
- Have it always be a value of `requestTimeout`, or maybe even 2x`requestTimeout`

nhtruong:
@ozonep Feel free to create a PR and thank you for the detailed explanation! ❤️

PR #678, opened 2023-12-21
Add expiration buffer to prevent credentials to expire earlier than request may finish.

ozonep:
### Description

Use 2x `requestTimeout` value as `expiration_buffer_ms` to make sure that whenever we make request credentials are valid for longer than request may be active and not time out (with added safety by using 2x value). 

### Issues Resolved

Fixes this: 
https://github.com/opensearch-project/opensearch-js/issues/677
### Check List

- [x ] New functionality includes testing.
  - [x ] All tests pass
- [ x] Linter check was successfull - `yarn run lint` doesn't show any errors
- [ x] Commits are signed per the DCO using --signoff
- [x ] Changelog was updated.

### DIFF
--- a/lib/aws/AwsSigv4Signer.js
+++ b/lib/aws/AwsSigv4Signer.js
@@ -14,6 +14,7 @@ const Transport = require('../Transport');
 const aws4 = require('aws4');
 const AwsSigv4SignerError = require('./errors');
 const crypto = require('crypto');
+const { toMs } = Transport.internals;
 
 const getAwsSDKCredentialsProvider = async () => {
   // First try V3
@@ -102,6 +103,9 @@ function AwsSigv4Signer(opts = {}) {
       }
 
       const currentCredentials = credentialsState.credentials;
+      // For AWS SDK V3 make sure token is valid at least until request has timed out.
+      const EXPIRATION_BUFFER_MS = toMs(options.requestTimeout || this.requestTimeout) * 2;
+
       let expired = false;
       if (!currentCredentials) {
         // Credentials haven't been acquired yet.
@@ -120,7 +124,10 @@ function AwsSigv4Signer(opts = {}) {
         expired = true;
       }
       // AWS SDK V3, Credentials.expiration is a Date object
-      else if (currentCredentials.expiration && currentCredentials.expiration < new Date()) {
+      else if (
+        currentCredentials.expiration &&
+        currentCredentials.expiration.getTime() - Date.now() < EXPIRATION_BUFFER_MS
+      ) {
         expired = true;
       }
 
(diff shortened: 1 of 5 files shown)


lib/aws/AwsSigv4Signer.js
```diff
@@ -102,6 +103,9 @@ function AwsSigv4Signer(opts = {}) {
       }
 
       const currentCredentials = credentialsState.credentials;
+      // For AWS SDK V3 make sure token is valid at least until request has timed out.
+      const EXPIRATION_BUFFER_MS = toMs(options.requestTimeout || this.requestTimeout) * 2;
```
dblock, line 107: It's not a constant, lowercase and maybe name it to something more specific, e.g. `credentialsExpireIn`?

dblock (commented):
This looks good. I have some asks below.

I think we need to document this in https://github.com/opensearch-project/opensearch-js/blob/main/USER_GUIDE.md#authenticate-with-amazon-opensearch-service at the very least, please?

Does it make sense to add an option that lets you configure this value per client instance? Thinking of this `*2` particularly.

If you feel super motivated we do want to extract authentication into a more detailed user guide similar to https://github.com/opensearch-project/opensearch-py/blob/main/guides/auth.md, so we could keep a short version in the main user guide and all the detailed options and options like this one in the detailed one.

lib/aws/AwsSigv4Signer.js
ozonep: I will lowercase it, but personally I'd leave `buffer` in its name, so it could be, e.g. `tokenExpirationBuffer`.
Because `buffer` gives instant understanding of what's the purpose of this variable, unlike `credentialsExpireIn`.

ozonep:
> This looks good. I have some asks below.
> 
> I think we need to document this in [`main`/USER_GUIDE.md#authenticate-with-amazon-opensearch-service](https://github.com/opensearch-project/opensearch-js/blob/main/USER_GUIDE.md?rgh-link-date=2023-12-21T15%3A28%3A28Z#authenticate-with-amazon-opensearch-service) at the very least, please?
> 
> Does it make sense to add an option that lets you configure this value per client instance? Thinking of this `*2` particularly.
> 
> If you feel super motivated we do want to extract authentication into a more detailed user guide similar to [opensearch-project/opensearch-py@`main`/guides/auth.md](https://github.com/opensearch-project/opensearch-py/blob/main/guides/auth.md?rgh-link-date=2023-12-21T15%3A28%3A28Z), so we could keep a short version in the main user guide and all the detailed options and options like this one in the detailed one.


I don't think it makes sense to allow altering this value - one can already do this by changing requestTimeout, plus that's the whole point - make sure credentials are valid until request is  made. 
I think I can remove 2x part - that was added simply for extra safety, I haven't tested 1x value (in production env), but should also be fine.

lib/aws/AwsSigv4Signer.js
dblock: A buffer usually a space/memory thing. But not a hill I want to die on :)

dblock (approved):

nhtruong (approved):
Looks great. Thank for making this :)

PR #758, opened 2024-04-17
adds a new export from the package for aws-sdk v3

### LINKED ISSUE #757: dynamic import bundles entire aws-sdk
When bundling opensearch-js with webpack or esbuild, the entire aws-sdk gets included in the bundle because of the import here https://github.com/opensearch-project/opensearch-js/blob/main/lib/aws/AwsSigv4Signer.js#L30C5-L30C43

This adds an additional 11MB to the final bundle size.

The code can be refactored to make it a static import that can be treeshook, maybe separate files for aws sdk v2 & aws sdk v3.

### What alternatives have you considered?
We've marked aws-sdk as external to keep it from bundling but that's risky as some other library might need aws-sdk

tusharf5:
### Description

Based on the conversation in this issue #757 , this PR adds a new way to import the AwsSigv4Signer as shown below.  

```
const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws-v3');
```

By using a new import path, the sdk remains backwards compatible while also adding a way for clients to avoid getting the entire aws-sdk included in the final bundle which results in an additional 22MB.

I tested the code and it does reduce the bundle size with the new import path

image

The bundle size is still not good though because '@aws-sdk/credential-provider-node' dynamically imports all types of credential providers ini, ec2, environment, node, process, http etc which is about 1MB in size. 

Would it make sense to have the client provide the provider based on the environment instead of relying on the default provider chain? This would be a breaking change but it would delegate the credential responsibility on the client eliminating the dependency on aws-sdk & @aws-sdk/credential-provider-node?

https://github.com/opensearch-project/opensearch-js/blob/main/lib/aws/AwsSigv4Signer.js#L74

We will have to throw an error on this (making it required) line so the client always provides a `getCredential` function and we can remove the dependency on a default provider (both v2 and v3) from our end.

The doc examples already use the `getCredential` function so I am guessing most of the clients would already be providing this function and not relying on the default `getCredential`.

https://opensearch.org/docs/latest/clients/javascript/index/#authenticating-with-amazon-opensearch-service-aws-signature-version-4

### Issues Resolved

Closes #757 

### Check List

  - [x] All tests pass
- [x] Changelog was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- A new import that only uses aws-sdk v3 for aws default credentials.
+```js
+const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws-v3');
+```
 ### Dependencies
 - Bumps `@types/node` from 20.12.5 to 20.12.7
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
-- A new import that only uses aws-sdk v3 for aws default credentials.
-```js
-const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws-v3');
-```
+- Added `@opensearch-project/opensearch/aws-v3` import ([758](https://github.com/opensearch-project/opensearch-js/pull/758)).
 ### Dependencies
 - Bumps `@types/node` from 20.12.5 to 20.12.7
 ### Changed
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -91,6 +91,8 @@ const client = new Client({
 const { defaultProvider } = require('@aws-sdk/credential-provider-node'); // V3 SDK.
 const { Client } = require('@opensearch-project/opensearch');
 const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws-v3'); // use aws-v3 import path if you are using aws-sdk v3
+// Unlike the import path in the v2 example above that lazy loads both aws-sdk v3 credential providers & entire aws-sdk v2 if available
+// This will only lazy load the aws-sdk v3 credential providers
 
 const client = new Client({
   ...AwsSigv4Signer({
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -90,7 +90,7 @@ const client = new Client({
 ```javascript
 const { defaultProvider } = require('@aws-sdk/credential-provider-node'); // V3 SDK.
 const { Client } = require('@opensearch-project/opensearch');
-const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws');
+const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws-v3'); // use aws-v3 import path if you are using aws-sdk v3
 
 const client = new Client({
   ...AwsSigv4Signer({
--- a/lib/aws/AwsSigv4Signer-sdk-v3.js
+++ b/lib/aws/AwsSigv4Signer-sdk-v3.js
@@ -20,7 +20,7 @@ const getAwsSDKCredentialsProvider = async () => {
     }
   } catch (err) {
     throw new AwsSigv4SignerError(
-      "Missing '@aws-sdk/credential-provider-node' module. Install it as a dependency"
+      "Missing '@aws-sdk/credential-provider-node' module. Install it as a dependency."
     );
   }
 };
--- a/lib/aws/AwsSigv4Signer-sdk-v3.js
+++ b/lib/aws/AwsSigv4Signer-sdk-v3.js
@@ -20,7 +20,7 @@ const getAwsSDKCredentialsProvider = async () => {
     }
   } catch (err) {
     throw new AwsSigv4SignerError(
-      'Unable to find a valid AWS SDK, please provide a valid getCredentials function to AwsSigv4Signer options.'
+      "Missing '@aws-sdk/credential-provider-node' module. Install it as a dependency"
     );
   }
 };
--- a/lib/aws/AwsSigv4Signer-sdk-v3.js
+++ b/lib/aws/AwsSigv4Signer-sdk-v3.js
@@ -13,36 +13,16 @@ const AwsSigv4SignerError = require('./errors');
 const { giveAwsV4Signer, loadAwsCredentialProvider } = require('./shared');
 
 const getAwsSDKCredentialsProvider = async () => {
-  // First try V3
   try {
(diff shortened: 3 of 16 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- A new import that only uses aws-sdk v3 for aws default credentials.
```
dblock, line 7: Match other CHANGELOG lines.

```
- Added `@opensearch-project/opensearch/aws-v3` import ([758](https://github.com/opensearch-project/opensearch-js/pull/758)).
```

lib/aws/AwsSigv4Signer-sdk-v3.js
```diff
@@ -0,0 +1,30 @@
...
+
+const getAwsSDKCredentialsProvider = async () => {
+  try {
+    const awsV3 = await import('@aws-sdk/credential-provider-node');
+    if (typeof awsV3.defaultProvider === 'function') {
+      return awsV3.defaultProvider();
+    }
+  } catch (err) {
+    throw new AwsSigv4SignerError(
+      'Unable to find a valid AWS SDK, please provide a valid getCredentials function to AwsSigv4Signer options.'
```
dblock, line 23: This error is difficult to action. I think a simple "Missing AWS SDK" error is fine. If you want to offer a hint, try something that one can actually do (e.g. add an import? a dependency?).

dblock (changes requested):
Good stuff. Document new functionality/changes in README, USERS_GUIDE, update samples. Add tests.

dblock:
> Would it make sense to have the client provide the provider based on the environment instead of relying on the default provider chain? This would be a breaking change but it would delegate the credential responsibility on the client eliminating the dependency on aws-sdk & @aws-sdk/credential-provider-node?

It's a good idea, but I would explore this in a future/separate PR.

tusharf5:
@dblock updated the PR with the requested changes

USER_GUIDE.md
```diff
@@ -90,7 +90,7 @@ const client = new Client({
 ```javascript
 const { defaultProvider } = require('@aws-sdk/credential-provider-node'); // V3 SDK.
 const { Client } = require('@opensearch-project/opensearch');
-const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws');
+const { AwsSigv4Signer } = require('@opensearch-project/opensearch/aws-v3'); // use aws-v3 import path if you are using aws-sdk v3
```
dblock, line 93: Should we say something about v2 here?

lib/aws/AwsSigv4Signer-sdk-v3.js
```diff
@@ -0,0 +1,30 @@
...
+
+const getAwsSDKCredentialsProvider = async () => {
+  try {
+    const awsV3 = await import('@aws-sdk/credential-provider-node');
+    if (typeof awsV3.defaultProvider === 'function') {
+      return awsV3.defaultProvider();
+    }
+  } catch (err) {
+    throw new AwsSigv4SignerError(
+      "Missing '@aws-sdk/credential-provider-node' module. Install it as a dependency"
```
dblock, line 23: Add a period at the end of the sentence?

dblock (approved):
👏  This looks great to me!

Let's get another pair of eyes, maybe @nhtruong and merge?

dblock:
@tusharf5 fix CI pls, there's some license headers missing, and a complaint about a new license, 0BSD. It's just a BSD variant so I think we can add it to the list, but I'll go double check.

USER_GUIDE.md
tusharf5: I've added some more comments here

dblock (approved):

dblock:
The license workflow needs to have 0BSD added to it, https://github.com/opensearch-project/opensearch-js/actions/runs/8739773216/job/23986128639?pr=758.

tusharf5:
@dblock fixed

tusharf5:
checking the test error

tusharf5:
1 existing test was failing since aws-sdk and v3 ask were not installed so the error thrown was different and now they are part of dev dependency so the error has changed. I've updated the test. hopefully no more test failures. @dblock

test/unit/lib/aws/awssigv4signer.test.js
```diff
@@ -11,6 +11,7 @@ const { test } = require('tap');
 const { URL } = require('url');
 const { v4: uuidv4 } = require('uuid');
 const AwsSigv4Signer = require('../../../../lib/aws/AwsSigv4Signer');
+const AwsSigv4SignerV3 = require('../../../../lib/aws/AwsSigv4Signer-sdk-v3');
```
nhtruong, line 14: Would it be easier and less confusing if we test this in a separate file/env?

nhtruong: Probably 4 different envs:
- `AwsSigv4Signer` with aws-v2
- `AwsSigv4Signer` with aws-v3
- `AwsSigv4SignerV3` with aws-v3
- `AwsSigv4SignerV3` without aws-v3

tusharf5: I've moved the tests to a different file

tusharf5:
Had to remove the peerDependency change since npm automatically installs peerDepenencies (yarn does not) and that changes the import result which fails the test. It should be working now.

test/unit/lib/aws/awssignv4signer-versions.test.js
```diff
@@ -0,0 +1,91 @@
...
+        t.same(
+          err.message,
+          "Missing '@aws-sdk/credential-provider-node' module. Install it as a dependency."
+        );
+      })
+      .finally(() => {
+        server.stop();
+      });
+  });
+});
```
nhtruong, line 91: Are these tests only testing if they throw the correct errors not for paths where no errors are thrown? I think that's why Codecov is not happy.

tusharf5: This would require mocking the import as the dependencies are not installed. While tap supports mocking imports, the project does not use the latest tap version which has breaking changes. 

Doesn't seem like there is any other way to test dependency missing/present scenarios without mocking.

dblock: We should solve this with integration tests IMO, which is a separate affair.

nhtruong: I'm thinking of setting up different `package.json` for each test case like you have in [test/bundlers/esbuild-test/package.json](https://github.com/opensearch-project/opensearch-js/pull/758/files/b0eb72e76c8f17490dfb381664dd06d2cb9ec2fe#diff-1cb683c49678d4831e23c772e985f2d08b5af71d8c0d3cc931399df98768069c)

dblock: It's a good idea. Do we want to hold this PR until then? Open an issue and do it later?

PR #823, opened 2024-07-22
Implement TransportRequestPromise, TransportRequestCallback in AWS Signer

### LINKED ISSUE #819: [BUG] AWS SIGv4 signer does not fully implement TransportRequestPromise
### What is the bug?

The `AwsSigv4SignerTransport.request`  method does not return a Promise that implements all of the TransportRequestPromise members.

https://github.com/opensearch-project/opensearch-js/blob/27c464faeb791fcc6aa1a17d73567212fbdd1432/lib/Transport.d.ts#L150-L153

https://github.com/opensearch-project/opensearch-js/blob/27c464faeb791fcc6aa1a17d73567212fbdd1432/lib/aws/shared.js#L80-L81

Create a new Client that uses the AWS SIGV4 signer.
Then attempting to call `.abort()` on the returned Promise of any of the client API methods will result in a TypeError being thrown.

The method should exist on the Promise returned by the Transport; in-flight requests that are SIGV4 signed should be abortable.

The 3.X fix is to accept an AbortSignal parameter on API methods which will allow in-flight requests to be cancelled.

u873838:
### Description

It was previously not possible to abort a request that used the AWS SigV4 signer transport.

### Issues Resolved

Fixes #819

### Check List

  - [X] All tests pass
- [X] Changelog was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Bumps `@types/node` from 20.14.10 to 20.14.11
 - Bumps `eslint-plugin-prettier` from 5.1.3 to 5.2.1
 ### Changed
+- Return a transport object from `AwsSigv4SignerTransport.request` that has an `.abort()` method that allows in-flight requests to be canceled
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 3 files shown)


dblock:
This looks good! Leaving to @nhtruong to do a thorough CR. In the meantime, maybe document this feature in a user guide?

nhtruong (approved):

nhtruong:
From what described in https://github.com/opensearch-project/opensearch-js/issues/819, this feature is expected/the-norm so we don't need to update the doc IMO.

PR #848, opened 2024-08-20
Replaced `docker-compose` with `docker compose` in Makefile

### LINKED ISSUE #845: [BUG] Workflows using Makefile are failing
Workflows using the Makefile to standup an OpenSearch cluster are [failing](https://github.com/opensearch-project/opensearch-js/actions/runs/10458956757/job/28961882949?pr=843):
>Run make cluster.clean cluster.opensearch.build cluster.opensearch.start
→ Cleaning up Docker assets...
docker volume prune --force
Total reclaimed space: 0B
docker network prune --force
docker system prune --volumes --force
Total reclaimed space: 0B
docker-compose --project-directory .ci/opensearch build;
/bin/sh: 1: docker-compose: not found
make: *** [Makefile:2: cluster.opensearch.build] Error 127
Error: Process completed with exit code 2.

nhtruong:
closes #845 

### DIFF
--- a/.github/workflows/integration-unreleased.yml
+++ b/.github/workflows/integration-unreleased.yml
@@ -48,8 +48,8 @@ jobs:
         uses: actions/setup-java@v3
         if: steps.cache-restore.outputs.cache-hit != 'true'
         with:
-          distribution: 'corretto'
-          java-version: ${{ matrix.entry.jdk_version }}
+          distribution: 'temurin'
+          java-version: ${{ matrix.entry.jdk_version || '17' }}
 
       - name: Assemble OpenSearch
         if: steps.cache-restore.outputs.cache-hit != 'true'
--- a/.github/workflows/integration-unreleased.yml
+++ b/.github/workflows/integration-unreleased.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Assemble OpenSearch
         if: steps.cache-restore.outputs.cache-hit != 'true'
         working-directory: opensearch
-        run: ./gradlew :distribution:archives:linux-tar:assemble
+        run: ./gradlew :distribution:archives:linux-tar:assemble --gradle-version=8.9
 
       - name: Save cached build
         if: steps.cache-restore.outputs.cache-hit != 'true'
@@ -72,7 +72,7 @@ jobs:
             fi
             echo '=====> waiting...'
           done
-          exit 1 
+          exit 1
 
       - name: Checkout Javascript Client
         uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


dblock (changes requested):
Let's not lose tests against 1.x. Add the JDK version to the matrix and install that one instead of 17.

https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/integration-unreleased.yml#L12

Makefile
```diff
@@ -1,12 +1,12 @@
 cluster.opensearch.build:
-	docker-compose --project-directory .ci/opensearch build;
+	docker compose --project-directory .ci/opensearch build;
 
 cluster.opensearch.start:
-	docker-compose --project-directory .ci/opensearch up -d ;
-	sleep 60;
+	docker compose --project-directory .ci/opensearch up -d ;
+	sleep 90;
```
dblock, line 6: Replace this with an active check? 

You can do a health check with https://github.com/opensearch-project/opensearch-benchmark/blob/f4ab3ab353ed2826e6c774f231878b387d2ab1c9/osbenchmark/resources/docker-compose.yml.j2#L41 or https://github.com/opensearch-project/dashboards-observability/blob/69d12cee1b0e43869d078e30cc996e9d685e44a9/docker-compose.yml#L47 maybe. Can be added in a future PR.

nhtruong:
> Let's not lose tests against 1.x. Add the JDK version to the matrix and install that one instead of 17.
Got it. Should we we remove that for 3.x client as its generated API no longer accept `doc_type` in OS 1.x?

dblock:
> > Let's not lose tests against 1.x. Add the JDK version to the matrix and install that one instead of 17.
> 
> @dblock Got it. Should we we remove that for 3.x client as its generated API no longer accept `doc_type` in OS 1.x?

If we do remove 1.x support then yes. I would take a hard look at the possibility of preserving 1.x support though because a very large number of users are still on 1.x and we would love for them to upgrade.

.github/workflows/integration-unreleased.yml
```diff
@@ -36,17 +36,25 @@ jobs:
...
         uses: actions/cache/restore@v3
         with:
           path: opensearch/distribution/archives/linux-tar/build/distributions
           key: ${{ steps.get-key.outputs.key }}
 
+      - name: Setup Java JDK
+        uses: actions/setup-java@v3
+        if: steps.cache-restore.outputs.cache-hit != 'true'
+        with:
+          distribution: 'corretto'
```
dblock, line 51: We use `temurin` everwhere else. 

You need a default, so `${{ matrix.entry.jdk_version || 17 }}`.

nhtruong:
@dblock all issues resolved. I also replaced `sleep 90` with an active health check script.

dblock (approved):

PR #864, opened 2024-09-04
Auto Generated API - 3.0.0

nhtruong:
### Changed
- All API functions are now generated from the OpenSearch API specification.
- API request and response types are now generated from the OpenSearch API specification.
- Overhauled API codebase and break it into smaller, more manageable files for better readability and maintainability.

### Removed
- *BREAKING* Removed support for API param aliases. That is, the API functions now only accept params with the exact names specified in the OpenSearch API specification.
- *BREAKING* Removed support for snake_cased API function names. All API functions are now camelCased only to conform to JavaScript naming conventions.
- *BREAKING* Removed support for Node.js 10 through 16. The minimum supported Node.js version is now 18.


dblock:
Let's branch 2.x before we merge this in order to have patch releases?

CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+### Dependencies
+### Changed
+- All API functions are now generated from the OpenSearch API specification.
+- API request and response types are now generated from the OpenSearch API specification.
+- Overhauled API codebase and break it into smaller, more manageable files for better readability and maintainability.
+### Deprecated
+### Removed
+- *BREAKING* Removed support for API param aliases. That is, the API functions now only accept params with the exact names specified in the OpenSearch API specification.
+- *BREAKING* Removed support for snake_cased API function names. All API functions are now camelCased only to conform to JavaScript naming conventions.
+- *BREAKING* Removed support for Node.js 10 through 16. The minimum supported Node.js version is now 18.
```
dblock, line 16: I think we need a more detailed UPGRADING.md and potentially remove these *BREAKING* warnings here like we do in other clients (e.g. https://github.com/opensearch-project/opensearch-py/blob/main/UPGRADING.md). For each breaking change users will appreciate a before/after example, but it's too much for a CHANGELOG.

dblock (commented):
It's looking good.

I don't know how much we care, but a very large amount of files are changed for a 1-liner in the license header. Adding an extra `*` would reduce this PR by half I imagine.

Screenshot 2024-09-05 at 7 04 01 AM

Another though on the camelcase/snake case. Having to update all code from one method to the other is going to be a barrier for users even though it's mostly a trivial change. Would it make sense to consider also generating wrapper methods with the old names and marking them deprecated for another major version iteration? Or to generate a compatibility layer that can be optionally included?

UPGRADING.md
```diff
@@ -0,0 +1,37 @@
+# Upgrading Opensearch JavaScript Client
+## Upgrading from 2.x to 3.x
+Starting from 3.x, the API functions and modules of the Opensearch JavaScript are generated from the OpenSearch REST API specification. While the vast majority of the API remains the same, there are some changes that you should be aware of when upgrading from 2.x to 3.x.
```
dblock, line 3:
```suggestion
Starting from 3.x, the API functions and modules in opensearch-js are generated from the [OpenSearch REST API specification](https://github.com/opensearch-project/opensearch-api-specification/). While the vast majority of the API remains the same, there are some breaking changes that may require updating your application when upgrading from 2.x to 3.x.
```

UPGRADING.md
```diff
@@ -0,0 +1,37 @@
...
+### Breaking Changes
+
+- **Removed camelCase params**: Every API function now only accepts parameters matching the OpenSearch API Spec, which are also what the OpenSearch server expects. In previous versions, some API functions would convert camelCase parameters to snake_case before sending them to the server. This is no longer the case. For example, the `client.search()` function now only accepts `filter_path`, and passing `filterPath` will result in an error from the server.
```
dblock, line 6:
```suggestion
#### Removed camelCase params

Every API function now only accepts parameters matching the OpenSearch API Spec, which are also what the OpenSearch server expects. In previous versions, some API functions would convert camelCase parameters to snake_case before sending them to the server. This is no longer the case. For example, the `client.search()` function now only accepts `filter_path`, and passing `filterPath` will result in an error from the server.
```

This is so one can link to the paragraph easily when it needs to be mentioned in, for example, bug reports.

UPGRADING.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+- **Removed overridden HTTP methods**: The 3.x client no longer allows overriding the HTTP method for API functions. In previous versions, some API functions let you pass `method` as a parameter to the API functions to override the default HTTP method. Now, `method` will be considered a querystring param, and using it will likely result in an error. If you need to send a custom request, you should use the `client.http` namespace.
```
dblock, line 9: Add a before/after example here for this one change.

UPGRADING.md
```diff
@@ -0,0 +1,37 @@
...
+const client = new Client({ node: 'http://localhost:9200' });
+
+const searchRequest: API.Search_Request = { body: { query: { match: { director: 'miller' } } } };
+const searchResponse = (await client.search(searchBody)).body; 
+// Note that the response type is now inferred, and you don't need to cast with `as API.Search_ResponseBody`
+```
+
+### Deprecations
+
+- **snake_case API modules and functions**: In 3.x, all API modules and functions are now in camelCase, comforming to the naming convention of Javascript. The snake_case aliases are still available but are deprecated and will be removed in a future version.
```
dblock, line 37:
```suggestion
#### snake_case API modules and functions

In 3.x, all API modules and functions are now in camelCase, conforming to the naming convention of Javascript. The snake_case aliases are still available but are deprecated and will be removed in a future version. For example, change all references to `create_pit` with `createPit`.
```

lib/pool/BaseConnectionPool.js
```diff
@@ -8,6 +8,7 @@
  *
  */
 
+
```
dblock, line 11: This is another instance of a change that causes hundreds of files to change, maybe not worth it.

dblock (commented):
Looking good. I tried to go through all the changes, made some suggestions in text.

I think after this is merged we should get another beta off main bofore releasing a final version. The final should be basically like the previous beta.

lib/pool/BaseConnectionPool.js
nhtruong: That was an accidental change due to GREP replace.
There are still gonna be nearly 800 file changes as we have nearly 400 API functions and each has a function definition and a type definition files.

nhtruong:
> Looking good. I tried to go through all the changes, made some suggestions in text.
> 
> I think after this is merged we should get another beta off main bofore releasing a final version. The final should be basically like the previous beta.

I agree. There will be 2 follow-up PRs after this one:
- After the `Search` endpoint has been fixed in the spec, which I'm working on, I'll run the generator again to apply that change among others changes since the last time the spec was pulled to generate this. We can make a last beta release after that.
- Another PR for the generator code itself, and the supporting workflows. This can also wait till 3.1.0 release.

dblock (approved):
Nice work. The fact that samples don't have any changes in them is very encouraging.

PR #879, opened 2024-09-26
Replace JSON11-produced hex escape codes with unicode escape codes

wjordan:
### Description

Since #784, `JSON11.stringify` produces JSON5 documents with hex escape codes (`\x1b`), which aren't standard JSON and cause `JSON.parse` to error. When using JSON11 code path, this PR applies a fix to replace all `\xXX` escape codes with the JSON-compatible equivalent Unicode escape codes (`\u00XX`).

This PR adds a unit test verifying the fix by ensuring a round-trip `s.deserialize(s.serialize(obj))` with an object containing long numerals and ANSI escape sequences completes successfully without error.

### Issues Resolved

Fixes opensearch-project/OpenSearch-Dashboards#7367.

### Check List

  - [ ] All tests pass
- [ ] Changelog was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Deprecated
 ### Removed
 ### Fixed
+- Replace JSON11-produced hex escape codes with unicode escape codes ([#879](https://github.com/opensearch-project/opensearch-js/pull/879))
 ### Security
 
 ## [3.0.0]
--- a/lib/Serializer.js
+++ b/lib/Serializer.js
@@ -37,6 +37,7 @@ const kJsonOptions = Symbol('secure json parse options');
 const JSON11 = require('json11');
 
 const isBigIntSupported = typeof BigInt !== 'undefined';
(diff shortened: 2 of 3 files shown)


nhtruong:
Thanks @wjordan 
Do you think this is a bug in JSON11 library?
Don't forget to update the CHANGELOG.md file.

wjordan:
> Do you think this is a bug in JSON11 library?

No, the json11 library writes JSON5 documents as expected, the bug is in the serialization code in this repo. See the added unit test which fails without this PR.

AMoo-Miki:
Ideally, this should be handled on JSON11. I will find how that is happening.

ycyr:
any update ?

nhtruong:
looks like JSON11 2.0 has already resolved this issue. Both 2.13.0 and 3.x are using JSON11 2.0. @ycyr are you having trouble with this? would you mind opening a ticket describing what's happening on your end?

ycyr:
@nhtruong  I've just upgrade my cluster to the version 2.18 (I was at the version 2.16) and the problem has been fixed.

nhtruong:
Closing this since this has already been resolved by JSON11 2.x

PR #913, opened 2024-11-15
[api_generator] API path builder update

nhtruong:
## Handling of more complex sets of URLs.

Previously, the api_generator assumed that all URL paths of the same `x-operation-group` shared the same static components (i.e. if you remove all `{path_param}` components in each path, the paths are identical):
For example these 4 paths of the `nodes.stats` group have an identical set of static components of `_nodes` and `stats`
- /_nodes/{node_id}/stats/{metric}/{index_metric}
- /_nodes/{node_id}/stats/{metric}
- /_nodes/{node_id}/stats
- /_nodes/stats

And the generator would build the path as
```ts
const path = ['/_nodes/', node_id, '/stats/', metric, '/', index_metric].filter(c => c).join('').replace('//', '/');
```
However this assumption not true for some operation groups like `cluster.stats`:
- /_cluster/stats/{metric}/{index_metric}/nodes/{node_id}
- /_cluster/stats/{metric}/nodes/{node_id}
- /_cluster/stats/nodes/{node_id}
- /_cluster/stats

Notice that the first 3 paths have `_cluster`, `stats` and `nodes` as their static components while the last one only has `_cluster` and `stats`. Therefore the following path that the generator use to make is incorrect:
```ts
const path = ['/_cluster/', metric, '/', index_metric, 'stats/nodes/', node_id].filter(c => c).join('').replace('//', '/');
```
because when none of the path params are provided, we would end up with `const path = '/_cluster/states/nodes'`, which is not a valid path. With this PR, the generator will be able to detect this edge case and generate a more complex path building strategy for `cluster.stats`:
```ts
let path;
if (metric != null && index_metric != null && node_id != null) {
  path = '/_cluster/stats/' + metric + '/' + index_metric + '/nodes/' + node_id;
}  else if (metric != null && node_id != null) {
  path = '/_cluster/stats/' + metric + '/nodes/' + node_id;
}  else if (node_id != null) {
  path = '/_cluster/stats/nodes/' + node_id;
}  else {
  path = '/_cluster/stats';
}
```
## Streamlined path building code
The generated path building logic for operation groups with uniform static path components is now simpler and more performant. Here are a few examples of old vs new code:
```ts
const path = ['/_nodes/', node_id, '/stats/', metric, '/', index_metric].filter(c => c).join('').replace('//', '/');
const path = ['/_nodes', node_id, 'stats', metric, index_metric].filter(c => c).join('/');
```
```ts
const path = ['/', index, '/_alias/', name].filter(c => c).join('').replace('//', '/');
const path = ['', index, '_alias', name].filter(c => c).join('/');
```
```ts
const path = ['/', index, '/_validate/query'].filter(c => c).join('').replace('//', '/');
const path = ['', index, '_validate/query'].filter(c => c).join('/');
```
---

### DIFF
--- a/api_generator/src/renderers/render_code/FunctionFileRenderer.ts
+++ b/api_generator/src/renderers/render_code/FunctionFileRenderer.ts
@@ -12,6 +12,7 @@ import BaseRenderer from '../BaseRenderer'
 import _ from 'lodash'
 import type ApiFunction from '../../spec_parser/ApiFunction'
 import type Namespace from '../../spec_parser/Namespace'
+import ApiPath from '../../spec_parser/ApiPath'
 
 export default class FunctionFileRenderer extends BaseRenderer {
   protected template_file = 'function.mustache'
@@ -35,8 +36,9 @@ export default class FunctionFileRenderer extends BaseRenderer {
       params_container_description: _.values(this.func.params).length === 0 ? ' - (Unused)' : undefined,
       parameter_descriptions: this.#parameter_descriptions(),
       function_name: this.func.function_name,
-      path_components: this.#path_components(),
-      path: this.#path(),
+      paths_are_uniform: ApiPath.statically_uniform(this.func.paths),
+      uniform_path: this.#uniform_path(),
+      diverged_paths: this.#diverged_paths(),
       http_verb: this.#http_verb(),
       body_required: this.func.request_body?.required,
       return_type: '{{abort: function(), then: function(), catch: function()}|Promise<never>|*}',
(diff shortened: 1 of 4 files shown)


dblock:
Why are these in the same operation group when they are different APIs?

```
/_cluster/stats/{metric}/{index_metric}/nodes/{node_id}
/_cluster/stats/{metric}/nodes/{node_id}
/_cluster/stats/nodes/{node_id}
/_cluster/stats
```

Should they be in separate operation groups?

#### cluster_stats_nodes

```
/_cluster/stats/{metric}/{index_metric}/nodes/{node_id}
/_cluster/stats/{metric}/nodes/{node_id}
/_cluster/stats/nodes/{node_id}
```

#### cluster_stats

```
/_cluster/stats
```

nhtruong:
@dblock that's how it's specified in the spec. The fix should be in the spec, if that's an error. Regardless, this code gen will now be able to handle this edge case if it appears in the spec.

nhtruong:
I dont think we should change the groups because of how it's implemented on a client. That will also result in a breaking change for all clients.

dblock (approved):
This is fine, but I bet in a year nobody will understand what it's doing for lack of tests. Approving to unblock.

PR #937, opened 2024-12-06
Fix document_lifecycle.md

### LINKED ISSUE #936: [BUG] Documentation error on document_lifecycle.md
### What is the bug?
The doc on [guides/document_lifecycle.md](https://github.com/opensearch-project/opensearch-js/blob/main/guides/document_lifecycle.md#setup) states that you can get the response of client.indices.exists as a boolean value but it's a http response, to get the boolean value you need to unpack it's "body" field.

Execute the code on guides/document_lifecycle.md, it will never create an index as the truth value of a filled object is true so the creation will never happen.

eusousu:
unpack exists boolean value from response body

### Description

Fix a mistake on docs

### Issues Resolved

Fixes #936 

### DIFF
--- a/guides/document_lifecycle.md
+++ b/guides/document_lifecycle.md
@@ -15,8 +15,8 @@ Next, create an index named `movies` with the default settings:
 ```javascript 
 const index = 'movies';
 client.indices.exists({ index: index })
-  .then({ body: exists } => {
-    if (!exists) {
+  .then(exists => {
+    if (!exists.body) {
       return client.indices.create({ index: index });
     }
   })
--- a/guides/document_lifecycle.md
+++ b/guides/document_lifecycle.md
@@ -15,7 +15,7 @@ Next, create an index named `movies` with the default settings:
 ```javascript 
 const index = 'movies';
 client.indices.exists({ index: index })
-  .then(exists => {
+  .then({ body: exists } => {
     if (!exists) {
       return client.indices.create({ index: index });
     }


guides/document_lifecycle.md
```diff
@@ -15,7 +15,7 @@ Next, create an index named `movies` with the default settings:
 ```javascript 
 const index = 'movies';
 client.indices.exists({ index: index })
-  .then(exists => {
+  .then({ body: exists } => {
```
nhtruong, line 18: This is not a proper syntax unfortunately.

guides/document_lifecycle.md
```diff
@@ -15,7 +15,7 @@ Next, create an index named `movies` with the default settings:
...
     if (!exists) {
```
nhtruong, line 19: I think we should simply access the body of the response (named exists) to get the boolean value:
`if(!exists.body) {`

nhtruong: Then the overall fix would be:
```
client.indices.exists({ index: index })
  .then(exists => {
  if (!exists.body) {
    return client.indices.create({ index: index });
  }
}).then(response => console.log(`Index ${response.body.index} created`));
```

nhtruong (approved):

guides/document_lifecycle.md
eusousu: Okay! Made the change :)

nhtruong (approved):

Issue #942, opened 2024-12-16
No typings on search body, other than Record<string, any>


paul-uz:
Is there a way to get better typings on the body of a search result for example?

```
  search<
    TResponse = Record<string, any>,
    TRequestBody extends RequestBody = Record<string, any>,
    TContext = Context
  >(
    params?: RequestParams.Search<TRequestBody>,
    options?: TransportRequestOptions
  ): TransportRequestPromise<ApiResponse<TResponse, TContext>>;
```

I see there is `SearchResponse` for Msearch, but not Search

nhtruong:
@paul-uz which version of the client are you using? We're testing Version 3.0.0 with [beta-8](https://www.npmjs.com/package/@opensearch-project/opensearch/v/3.0.0-beta.8) already out that have all request and response types.

paul-uz:
Just using the latest stable release, set as ^2.13.0 in our package file. When will V3 be released officially?

paul-uz:
I just installed beta-8, and whilst i can click through and find the new type ie

```
export type Search_Response = ApiResponse & {
  body: Search_ResponseBody;
}
```

`body` is just showing as `any` in my IDE, VSCode :/

paul-uz:
VSCode intellisense gives me this typing

```
const response: {
    body: any;
    statusCode: number | null;
    headers: Record<string, any> | null;
    warnings: string[] | null;
    meta: {
        context: unknown;
        name: string | symbol;
        request: {
            params: TransportRequestParams;
            options: TransportRequestOptions;
            id: any;
        };
        connection: Connection;
        attempts: number;
        aborted: boolean;
        sniff?: {
            hosts: any[];
            reason: string;
        } | undefined;
    };
    body: ResponseBody;
}
```

`body` is deifned twice somehow

nhtruong:
The typing in this [comment](https://github.com/opensearch-project/opensearch-js/issues/942#issuecomment-2547150756) is coming from the 2.x version. Make sure that the 2.x package has been removed from your env.

paul-uz:
> The typing in this [comment](https://github.com/opensearch-project/opensearch-js/issues/942#issuecomment-2547150756) is coming from the 2.x version. Make sure that the 2.x package has been removed from your env.

I simply installed the beta over the top via NPM.

I'll try removing node_modules and reinstalling.

paul-uz:
Code from 3.0.0-beta.8

```
export type Search_Response = ApiResponse & {
[...]

PR #943, opened 2024-12-18
fix: revert Search_Response

paul-uz:
Reverted a change between `beta.7` and `beta.8` that broke the Search_Response typings.

Closes https://github.com/opensearch-project/opensearch-js/issues/942

### DIFF
--- a/api/_core/search.d.ts
+++ b/api/_core/search.d.ts
@@ -106,9 +106,8 @@ export type Search_RequestBody = {
   version?: boolean;
 }
 
-export type Search_Response = ApiResponse & {
+export interface Search_Response extends ApiResponse {
   body: Search_ResponseBody;
 }
 
 export type Search_ResponseBody = Core_Search.ResponseBody
-


nhtruong:
Thanks. That file is actually generated via the [api_generator](https://github.com/opensearch-project/opensearch-js/tree/main/api_generator). So changes must be made in the generator and then regenerate the API

paul-uz:
> Thanks. That file is actually generated via the [api_generator](https://github.com/opensearch-project/opensearch-js/tree/main/api_generator). So changes must be made in the generator and then regenerate the API

Would you be able to point me in the right direction on how to do so? Ta

nhtruong:
[Here](https://github.com/opensearch-project/opensearch-js/blob/main/api_generator/src/renderers/render_types/TypesFileRenderder.ts#L73-L92) is where most of the changes will have to be made but it's not easy. My memory on this is till pretty fresh so I'll tackle this.

PR #1018, opened 2025-05-14
Add Error Listener Back Before Calback 

sejli:
### Description

Thanks @ananzh for the following investigation.

In the `onResponse()` function in `Connection.js`, `cleanListeners()` removes all event listeners from the request, including the error listener.

```
function cleanListeners() {
  request.removeListener('response', onResponse);
  request.removeListener('timeout', onTimeout);
  request.removeListener('error', onError);
  request.removeListener('abort', onAbort);
  cleanedListeners = true;
}
```
After the response is handled and the listeners are removed, the remainder of the request body tries to write to the now-closed connection
This triggers an error event on the request object
Since there are no error listeners attached to the request (they were removed by cleanListeners()), this becomes an uncaught exception that crashes the Node.js process

By adding the error listener back, we are able to keep an error listener on the request in the event that the response comes back before the entire request is sent.

### Issues Resolved

### Check List

  - [ ] All tests pass
- [ ] Changelog was updated.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -112,6 +112,8 @@ class Connection {
     const onResponse = (response) => {
       cleanListeners();
       this._openRequests--;
+      // re-add an error listener
+      request.addListener('error', (err) => {console.log('ignoring error after response', err)});
       callback(null, response);
     };
 


nhtruong:
Thanks @sejli

nhtruong (approved):

Issue #1098, opened 2026-04-23
[BUG] x-amz-content-sha256 header is unsigned for aws SigV4 requests


Anqiwa-dev:
### What is the bug?
_A clear and concise description of the bug._

The x-amz-content-sha256 header was set on the request after aws4.sign() was called in 
`shared.js`. This meant the header was present on the wire but not included in the SignedHeaders list of the Authorization header. 

With the Amazon OpenSearch Serverless (AOSS) NextGen launch, it will require `x-amz-content-sha256` to be signed, following the AWS recommendation that all headers with `x-amz-*` format should be signed. 

Reference - https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
> "You must include the host header and any x-amz-* headers in the signature."

### How can one reproduce the bug?
_Steps to reproduce the behavior._

1. Create a Opensearch Serverless Collection with NextGen endpoint
2. Send a request with non-empty body
3. Will receive response from Opensearch Serverless NextGen with 400 error message
```
`x-amz-content-sha256 header must be signed in SigV4`
```

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

x-amz-content-sha256 header should be signed.

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

No

### Do you have any additional context?
_Add any other context about the problem._

No

krisfreedain:
@ananzh @kavilla @VachaShah maintainers - please take a look

[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross) [3](https://github.com/sean-) [4](https://github.com/Ritesh9876)

PR #1099, opened 2026-04-23
Fix x-amz-content-sha256 header not being signed in SigV4

Anqiwa-dev:
### Description

_Describe what this change achieves._

Fix `x-amz-content-sha256` header not being included in SigV4 SignedHeaders when using AwsSigv4Signer.

Previously, the x-amz-content-sha256 header was set on the request after aws4.sign() was called in 
`shared.js`. This meant the header was present on the wire but not included in the SignedHeaders list of the Authorization header. 

In OpenSearch Serverless, it will require `x-amz-content-sha256` to be signed, following the AWS recommendation that all headers with `x-amz-*` format should be signed. 

Reference - https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
> "You must include the host header and any x-amz-* headers in the signature."

Without this fix, the request to Opensearch Serverless would return  a 400 error: `x-amz-content-sha256 header must be signed in SigV4`.

The fix moves the x-amz-content-sha256 assignment before the aws4.sign() call so it is included in the signature computation.

### Issues Resolved

https://github.com/opensearch-project/opensearch-js/issues/1098

### Check List

  - [X] All tests pass
- [] Linter check was successfull - `yarn run lint` doesn't show any errors 
    - errors are unrelated to my change 
- [X] Changelog was updated.

### DIFF
--- a/lib/aws/shared.js
+++ b/lib/aws/shared.js
@@ -65,11 +65,12 @@ function giveAwsV4Signer(awsDefaultCredentialsProvider) {
         request.service = awssigv4Cred.service;
         delete request['auth'];
       }
-      const signed = aws4.sign(request, credentialsState.credentials);
-      signed.headers['x-amz-content-sha256'] = crypto
+
+      request.headers['x-amz-content-sha256'] = crypto
         .createHash('sha256')
         .update(request.body || '', 'utf8')
         .digest('hex');
+      const signed = aws4.sign(request, credentialsState.credentials);
       return signed;
     }
 
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -21,10 +21,12 @@ jobs:
         entry:
           - { opensearch_version: 1.1.0 }
           - { opensearch_version: 1.2.4 }
+          - { opensearch_version: 1.3.19 }
           - { opensearch_version: 1.3.20 }
-          - { opensearch_version: 2.3.0 }
+          - { opensearch_version: 2.2.1 }
           - { opensearch_version: 2.6.0 }
           - { opensearch_version: 2.12.0 }
+          - { opensearch_version: 2.16.0 }
           - { opensearch_version: 2.17.0 }
           - { opensearch_version: 2.19.0 }
     env:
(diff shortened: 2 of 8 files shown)


rbhatane (approved):

harshavamsi (approved):

lib/aws/shared.js
```diff
@@ -65,11 +65,12 @@ function giveAwsV4Signer(awsDefaultCredentialsProvider) {
         request.service = awssigv4Cred.service;
         delete request['auth'];
       }
-      const signed = aws4.sign(request, credentialsState.credentials);
-      signed.headers['x-amz-content-sha256'] = crypto
+
+      request.headers['x-amz-content-sha256'] = crypto
```
seraphjiang, line 69: is this field optional? what if request doesn't has this?

Anqiwa-dev: This is calculating the body checksum for the request and set it for the request. This header should not optional - https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html


> The x-amz-content-sha256 header is required for all AWS Signature Version 4 requests.

seraphjiang: Thanks @Anqiwa-dev .  why do we need this header now not before.
 have we test existing endpoint to see the result. if we tested, could we share the result

ananzh (approved):

lib/aws/shared.js
ananzh: @Anqiwa-dev I think this change is for MDS so we connect to remote OpenSearch clusters via SigV4 authentication right?

ananzh: I think the unit test can show the header is updated but not show whether the serverless endpoint accept the newly-signed request end-to-end. If you have time, I think we could change package.json and point to local js client                                                                                                                                                                            
                                                                                                                                                                                                                                  "@opensearch-project/opensearch": "file:../opensearch-js-patched"

then hit with a real Serverless endpoint with result before and after.

Anqiwa-dev: > I think the unit test can show the header is updated but not show whether the serverless endpoint accept the newly-signed request end-to-end. If you have time, I think we could change package.json and point to local js client

Thank you for the suggestion. It make sense. However, a real http call should be part of an integration test instead of unit test right? Would adding this additional dependency in a unit test make it to be more flaky? I will add a CI flow after the new serverless endpoint GA (right now it should not publicly available)

seraphjiang (approved):

peterzhuamazon (approved):

PR #1146, opened 2026-09-18
Declare the Connection agent property as agent, not _agent

### LINKED ISSUE #443: [BUG][CCI] Fixed typo in connection `_agent`

gohri81-tech:
### Description

Closes #443.

`lib/Connection.d.ts` declares

```ts
_agent: http.Agent | https.Agent | hpagent.HttpProxyAgent | hpagent.HttpsProxyAgent;
```

but `lib/Connection.js` never assigns `this._agent`. The constructor assigns `this.agent` in all four branches (the `agent` function case, the `agent: false` case, the proxy case and the plain case), and `close()` reads it back:

```js
if (this.agent !== undefined) {
  this.agent.destroy();
}
```

So the type declares a property that does not exist at runtime, and omits the one that does. `_agent` in `lib/pool/index.d.ts` is a different thing and is correct - `BaseConnectionPool` really does set `this._agent = opts.agent`.

This renames the declaration to `agent`, which is the direction @timursaurus suggested on #442: the property is not internal, so it should not carry the underscore that the genuinely internal `_openRequests` and `_status` do (both of which do match `Connection.js`).

### What was broken

@dblock asked this on #442, so to be concrete - the bug is types-only, but it is not invisible:

- A TypeScript consumer who reaches for the agent gets `Property 'agent' does not exist on type 'Connection'`, which is wrong. Reading the agent is a reasonable thing to do; it is how you would check the socket settings a custom `agent` function produced, or call `destroy()` on it yourself.
- The reverse is worse: `connection._agent` compiles cleanly and is `undefined` at runtime, so the type actively leads you into a bug rather than catching it.

Because the declaration file has no runtime behaviour, there is nothing a Jest test can assert here - the existing suite passes both before and after. A compile-time assertion would be the right kind of test, but I could not find a `tsd`/`*.test-d.ts` harness in the repository to add one to. If you would like one, say so and I will add the harness and an assertion in this PR.

### Not in this PR

`close()` guards with `this.agent !== undefined`, and the constructor does set it to `undefined` when `opts.agent === false`, so strictly the type should be `... | undefined`. I have left that out because it would be a breaking change for anyone with `strictNullChecks` on who currently uses the property without a guard, and the issue is about the name. Happy to add it here or separately if you would rather have the type be exactly right.

I also have not added a CHANGELOG entry, following @dblock's note on #442 that changelogs can be skipped for trivial things. Tell me if you would prefer one and I will add it under `[Unreleased]` / `Fixed` with this PR's number.

### Issues Resolved

Closes #443

### Check List

- [x] New functionality includes testing: not applicable, see above - declaration-only change
- [x] Commits are signed per the DCO using `--signoff`

I was not able to run the build or `tsc` in my environment, so please treat CI as the source of truth on this one. The change is one line, from reading `lib/Connection.js` against `lib/Connection.d.ts`.


gohri81-tech:
A gentle reminder that this one is still waiting for a first look, whenever a maintainer has a moment. It is a small type declaration fix - the `Connection` class exposes the property as `agent`, so the declaration should say `agent` rather than `_agent`.

Happy to add a test or adjust anything if you would like it shaped differently. Thank you.

harshavamsi (approved):

